PR #4, opened 2023-11-10
Added testing section

michael-fielding:
Added testing documentation section (scaffold)


documentation/testing/images/path-to-live.jpg
Chris-Lorch, line 1: Please add in peer approval between "merge to main" 

Also, should it be "merge to dev branch" not "merge to main branch"?

documentation/testing/images/testing-pyramid.jpg
Chris-Lorch, line 1: Based on the languages and frameworks we are planning to use, do we have more targeted testing services we could apply? Can we also specify which tests are used in the service pipeline, vs the etl pipelines?

documentation/testing/images/path-to-live.jpg
michael-fielding: That is not our branching strategy. There is no need for a dev or feature sub branch

michael-fielding: And a PR is peer approval

documentation/testing/images/testing-pyramid.jpg
michael-fielding: This is the standard boilerplate we start with. I have just added this as the scaffold, and I expect @Faizi52996 to document the testing strategies here

documentation/testing/images/path-to-live.jpg
joekadi, line 1: few thoughts here:

1. seems a bad practice to wait code is deployed to prod to run smoke tests? should run all tests on the earlier envs so no need to do any testing on prod. 
2. we should have devX environments that we can deploy feature branches too to test on before merging to develop. 
3. i think we should run all tests on every env: devX->dev->test->preprod (i guess maybe just save the NFT tests till preprod as i assume the earlier envs will have less resources so won't have capacity to handle NFT tests)
4. seems best to have a code branch that deploys to each env, this diagram doesn't detail that. 
5. very trivial but weird to call the branch that deploys to dev env "main", better following standards and calling it "develop"

documentation/testing/images/types-of-testing.jpg
Chris-Lorch, line 1: Please label what the changes are in the y axis here, I assume it goes from automated to manual but cannot be sure.

documentation/testing/images/path-to-live.jpg
michael-fielding: All this has been covered in the branching strategy meting. In short, we will be using trunk-based development rather than Gitflow.

Faizi52996 (approved):

PR #14, opened 2023-12-05
Add terraform for core infrastructure and build / deployment pipeline

jrabbott:
1.  Add core infrastructure terraform (App Insights & Key Vault)
2.  Manual pipeline (with no triggers, etc) for initial setup - will require devops library to pass parameter to pipeline


Faizi52996 (approved):
had an overview. looks good to me.

PR #28, opened 2023-12-11
Feature/186502/yml file

spicy-rice:
* created a yaml file in pipeline, containing scripts


jrabbott (changes requested):
Please rebase from master, so that we can just review the required changes.

PR #29, opened 2023-12-11
Adds school expenditure end point

jrabbott:
Adds back-end school expenditure end point 

[AB#187930](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/187930)


michael-fielding (commented):
Overall it looks good. However, you have many public API methods without any XML comments, and several with just empty ones.

michael-fielding (approved):

PR #31, opened 2023-12-11
yaml file 

spicy-rice:
- created in /pipelines/

### DIFF
--- a/front-end-components/pipelines/front-end.yaml
+++ b/front-end-components/pipelines/front-end.yaml
@@ -1,31 +1,49 @@
 trigger:
-- main
-- feature/*
+  batch: true
+  branches:
+    include:
+    - main
+  paths:
+    include:
+    - front-end-components
+
 
 pr:
-- main
-- feature/*
+- paths:
+    include:
+    - front-end-components
+  drafts: false
 
 pool:
-  vmImage: ''  
+  vmImage: 'windows-2022'  
 
-steps:
-- checkout: self
-  fetchDepth: 1 
 
 - task: Npm@1
   displayName: 'Install dependencies'
   inputs:
     command: 'install'
+  workingDir: 'front-end-components'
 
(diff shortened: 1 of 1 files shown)


front-end-components/pipelines/front-end.yaml
```diff
@@ -0,0 +1,31 @@
+trigger:
+- main
+- feature/*
```
michael-fielding, line 3: Don't think we want to trigger whenever you push to your feature branch. It should be when a change is made to main only

front-end-components/pipelines/front-end.yaml
```diff
@@ -0,0 +1,31 @@
...
+  displayName: 'Run tests'
+  inputs:
+    command: 'custom'
+    customCommand: 'test'
+
+- task: Npm@1
+  displayName: 'Build project'
+  inputs:
+    command: 'custom'
+    customCommand: 'run build'
```
michael-fielding, line 31: Where is the step to deploy to the artifact pipeline?

front-end-components/pipelines/front-end.yaml
jrabbott, line 1: trigger section needs updating to the following;
image

pr section needs updating to the following;
image

pool vm image should be 'windows-2022'

Add the following to each of the NPM steps;

`workingDir: 'front-end-components'`

Checkout step isn't required, therefore can be removed.

jrabbott (changes requested):

jrabbott (changes requested):
Need to review spacing / indentation. Has to be correct or the pipeline will fail.

jrabbott (approved):

front-end-components/pipelines/front-end.yaml
jrabbott: @michael-fielding I think this is resolved now?

PR #35, opened 2023-12-13
Adds horizontal bar chart component, configures the react GDS component library and fixes jest configuration to work with css imports

joekadi:


front-end-components/src/components/HorizontalBarChart/HorizontalBarChart.tsx
```diff
@@ -0,0 +1,125 @@
...
+              return {
+                  weight: weight,
+              };
+            }
+          }
+        }
+      }
+    };
+    return(
+      <>
```
jrabbott, line 117: Are the empty elements needed?

joekadi: they are not! leftover. removed

jrabbott (approved):

PR #37, opened 2023-12-14
Adds school end point to establishment api

jrabbott:
Adds school end point to establishment api


platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Cosmos/CollectionService.cs
```diff
@@ -0,0 +1,50 @@
...
+        var groups = await GetAllActiveCollections();
+        return groups.Where(c => c.DataGroup == dataGroup)
+            .ToArray();
+    }
+
+    public async Task<DataCollection> GetLatestCollection(string dataGroup)
+    {
+        var activeCollections = await GetActiveCollectionsForDataGroup(dataGroup);
+
+        return activeCollections.MaxBy(o => o.Name.Split('-').First())
```
Chris-Lorch, line 47: If _Name_ is a string, e.g. "2022-2023" would the _MaxBy_ function still work? I'd assume that the call to _Split_ would return "2022","2023" as string types, not numeric?

jrabbott: The naming convention on the collections is  {timestamp}-{type}-{year_from}-{year_to}. 

The spilt is getting the first element which is the timestamp. 

MaxBy function isn't limited to use numberic. It will use build-in comparer (or custom should you so wish) for the type building used.

https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.maxby?view=net-8.0

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Cosmos/CosmosClientFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+        return new CosmosClient(connectionString, new CosmosClientOptions
+        {
+#if DEBUG
+            //Disabling SSL validation for local emulator
+            HttpClientFactory = () =>
+            {
+                HttpMessageHandler httpMessageHandler = new HttpClientHandler
+                {
+                    ServerCertificateCustomValidationCallback =
+                        HttpClientHandler.DangerousAcceptAnyServerCertificateValidator
```
Chris-Lorch, line 18: My understanding of `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` is that it allows you to use custom certificate verification, and while I can see the use for this for dev and test environments, it looks like it will introduce an area of risk in production. So, should we replace this with a different method to remove any risk from the beginning?

jrabbott: `#if DEBUG ` is this a complier directive, such that when we release build (for dev, test and live) this code is not even present.

The CI/CD pipeline builds the code in release mode. The only time we'd you debug is locally.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Cosmos/DataGroups.cs
```diff
@@ -0,0 +1,14 @@
+namespace EducationBenchmarking.Platform.Infrastructure.Cosmos;
+
+public static class DataGroups
+{
+    public const string Edubase = "Edubase";
```
Chris-Lorch, line 5: Edubase is the older, redundant name for what the clients have now began calling "GIAS", which may cause confusion to devs without domain knowledge. Should we replace this before it is propagated it into the new service?

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Search/SearchOptions.cs
```diff
@@ -9,19 +9,4 @@ public abstract class SearchServiceOptions
...
-
-public class SchoolSearchServiceOptions : SearchServiceOptions
-{
-    public string SchoolIndexName { get; set; }
-}
-
-public class LocalAuthoritySearchServiceOptions : SearchServiceOptions
-{
-    public string LocalAuthorityIndexName { get; set; }
 }
```
Chris-Lorch, line 12: Has this been removed as it has been abstracted in the datacollection class?

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Cosmos/DataGroups.cs
jrabbott: Something we need to consider, if Edubase is a simple mirror of GIAS then it would make sense to change this.

Code impact is minimal, however we'd need to remove the fibre collection before committing to this change.

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/EducationBenchmarking.Platform.Infrastructure.csproj
```diff
@@ -9,7 +9,9 @@
 
     <ItemGroup>
       <PackageReference Include="Azure.Search.Documents" Version="11.5.1" />
-      <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.0" />
+      <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.37.0" />
+      <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
+      <PackageReference Include="System.Linq.Async" Version="6.0.1" />
     </ItemGroup>
 
 </Project>
```
Chris-Lorch, line 17: Should we be looking for the latest version for a particular release rather than keeping them fixed? e.g. 

`<PackageReference Include="Azure.Search.Documents" Version="11.*" />` 
vs
`<PackageReference Include="Azure.Search.Documents" Version="11.5.1" />1

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/Search/SearchOptions.cs
Chris-Lorch: nevermind! Found them.

jrabbott: No, I just put it at the wrong level.

platform/src/abstractions/EducationBenchmarking.Platform.Infrastructure/EducationBenchmarking.Platform.Infrastructure.csproj
jrabbott: Constantly using latest can cause instability in the code base. 

We should look to pin versions where possible, and manager version updates as and when needed.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/Edubase.cs
```diff
@@ -0,0 +1,132 @@
...
+    
+    public class LocationDataObject
+    {
+        [JsonProperty(PropertyName = FieldNames.LOCATION_TYPE)]
+        public string type { get; set; }
+
+        [JsonProperty(PropertyName = FieldNames.LOCATION_COORDINATES)]
+        public string[] coordinates { get; set; }
+    }
+}
```
Chris-Lorch, line 132: Is the naming convention used for the _FieldNames_ attributes standard or personal preference? If the latter then I think we should have an agreed attribute naming convention used within the team to keep everything consistent.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/FieldNames.cs
```diff
@@ -0,0 +1,53 @@
...
+        public const string LOCATION = "Location";
+        public const string TRUSTS = "Trusts";
+        public const string SPONSORS = "SchoolSponsors";
+        public const string LA_CODE = "LACode";
+        public const string ESTAB_NO = "EstablishmentNumber";
+        public const string LA_ESTAB = "LAEstab";
+        public const string TEL_NO = "TelephoneNum";
+        public const string NO_PUPIL = "NumberOfPupils";
+        public const string STAT_LOW = "StatutoryLowAge";
+        public const string STAT_HIGH = "StatutoryHighAge";
```
Chris-Lorch, line 24: Is STAT_LOW and STAT_HIGH intuitive enough for someone to know it relates to age?

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/FieldNames.cs
```diff
@@ -0,0 +1,53 @@
...
+        public const string MAT_SAT = "MAT SAT or Central Services";
+        public const string IS_FEDERATION = "IsFederation";
+        public const string FEDERATION_MEMBERS = "FederationMembers";
+        public const string IS_PART_OF_FEDERATION = "IsPartOfFederation";
+        public const string FEDERATION_UID = "FederationUId";
+        public const string FEDERATION_NAME = "FederationName";
+        public const string FEDERATIONS_CODE = "FederationsCode";
+        public const string FEDERATION = "Federations";
+        public const string GOV_OFFICE_REGION = "GOR";
+}
```
Chris-Lorch, line 53: Do we need to enforce data types at this stage? e.g. _PERIOD_COVERED_BY_RETURN_ is listed as type string, but is better suited as an integer datatype.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/Edubase.cs
jrabbott: Naming convention should follow the language specific code styles. However actually naming on variables, methods, etc should be meaning without being overly verbose.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/FieldNames.cs
jrabbott: Those are field name constants, not data types for the field it's self. As such the type is string.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/Edubase.cs
```diff
@@ -0,0 +1,132 @@
+using Newtonsoft.Json;
+
+namespace EducationBenchmarking.Platform.Api.Establishment.Db;
+
+public class Edubase
```
Chris-Lorch, line 5: Why is this named Edubase? Does it relate to the GIAS/edubase data, or is it going to be populated with information from the maintained / academies reports? 

I'm trying to understand the relevance of calling it Edubase here, and whether it might cause confusion with something else.

jrabbott: It relates to the Edubase dataset/collection.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Search/LocalAuthoritySearchService.cs
```diff
@@ -0,0 +1,31 @@
...
+    public LocalAuthoritySearchService(IOptions<LocalAuthoritySearchServiceOptions> options) : base(options.Value.Endpoint, options.Value.LocalAuthorityIndexName, options.Value.Credential)
+    {
+    }
+
+    public async Task<SearchOutput<LocalAuthority>> SearchAsync(PostSearchRequest request)
+    {
+        return await SearchAsync<LocalAuthority>(request, CreateFilterExpression, Facets);
+    }
+
+    private string CreateFilterExpression(List<SearchFilters> requestFilters)
```
Chris-Lorch, line 27: Just a general question, but is this is a placeholder for now, and would we expect to fill it out?

Chris-Lorch (approved):
As a general comment, I think we could do with more logging, as this is something both current services lack and has hampered our ability to pick up and decipher the services. 

I'm also someone who loves comments in code. For me as a dev, if I'm unfamiliar with the code it helps me quickly understand components in the code at a glance, which I can then go back and digest in more detail. I'd imagine I'm not the only one like this, and it could help other devs when they pick up tickets to work on part of the code base they've never seen before.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Search/LocalAuthoritySearchService.cs
jrabbott: Yes - however it isn't a placeholder, it's intentional as there isn't any knew filters for local authorities. By put this pattern in place it will allow you to see where to add the implementation as and when we know more.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/Db/FieldNames.cs
Chris-Lorch: Is this something we can do now? Or not yet since we are still using the SFB datasets?

jrabbott: My preference for these field name constants is to keep them as is for now (just because of the serialisation issues and badly named fields). However, everything else most definitely.

PR #38, opened 2023-12-15
fixed indentations

spicy-rice:


front-end-components/pipelines/front-end.yaml
jrabbott, line 1: Gone the wrong way with the indentation :-(

jrabbott (approved):

PR #75, opened 2024-01-12
Testing Pipeline decision document

Faizi52996:
The decision regarding our testing pipeline strategy will significantly impact the efficiency and effectiveness of our development process. I propose that we review these options collectively, considering our project's specific needs, infrastructure capabilities, and desired testing outcomes.
Please feel free to provide your feedback and preferences, and let's collaborate to arrive at a consensus on the most suitable approach for our testing pipelines.

### DIFF
--- a/documentation/testing/Decisions/Testing-Pipeline-SetupOptions.md
+++ b/documentation/testing/Decisions/Testing-Pipeline-SetupOptions.md
@@ -1,5 +1,5 @@
 ﻿# Background
-Our projects involve two main solutions - `EducationBenchmarking.Platform` and `EducationBenchmarking.Web`. Each solution has its unique set of tests: unit tests, integration tests, functional API tests for `Platform`, and Feature/E2E tests along with accessibility tests for `Web`. The focus of our exploration is on optimizing the testing pipelines for the functional API tests for `Platform` and Feature/E2E tests and accessibility tests for `Web`.
+Our projects involve two main solutions - `EducationBenchmarking.Platform` and `EducationBenchmarking.Web`. Each solution has its unique set of tests: unit tests, integration tests, functional API tests for `Platform`, and Feature/E2E tests along with accessibility tests for `Web`. The focus of our exploration is on optimising the testing pipelines for the functional API tests for `Platform` and Feature/E2E tests and accessibility tests for `Web`.
 
 ## Context and Problem Statement
 In the current state of our development lifecycle, we find ourselves at a stage where the absence of established testing pipelines demands our attention. To fortify the reliability and effectiveness of our projects, we recognize the essential need to implement robust testing structures. We need a solution that fulfills the following criteria.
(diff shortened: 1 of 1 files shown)


documentation/testing/Decisions/Testing-Pipeline-SetupOptions.md
```diff
@@ -0,0 +1,41 @@
+﻿# Background
+Our projects involve two main solutions - **EducationBenchmarking.Platform** and **EducationBenchmarking.Web**. Each solution has its unique set of tests: functional API tests for Platform and Feature/E2E tests along with accessibility tests for Web. The focus of our exploration is on optimizing the testing pipelines for these specific test suites.
```
Chris-Lorch, line 2: Check Americanised spelling, e.g. "optimizing vs optimising", "minimize vs minimise"

Chris-Lorch (commented):
added my thoughts

documentation/testing/Decisions/Testing-Pipeline-SetupOptions.md
```diff
@@ -0,0 +1,50 @@
+﻿# Background
+Our projects involve two main solutions - `EducationBenchmarking.Platform` and `EducationBenchmarking.Web`. Each solution has its unique set of tests: unit tests, integration tests, functional API tests for `Platform`, and Feature/E2E tests along with accessibility tests for `Web`. The focus of our exploration is on optimising the testing pipelines for the functional API tests for `Platform` and Feature/E2E tests and accessibility tests for `Web`.
+
+## Context and Problem Statement
+In the current state of our development lifecycle, we find ourselves at a stage where the absence of established testing pipelines demands our attention. To fortify the reliability and effectiveness of our projects, we recognize the essential need to implement robust testing structures. We need a solution that fulfills the following criteria.
+
+### Criteria
+1. **Simple Tests:** We want to keep the tests as simple as possible to make them easily manageable.
+2. **Cost-Effective:** The solution we implement should be mindful of resource allocation and ensure cost-effectiveness without compromising on the quality of our testing processes.
```
Chris-Lorch, line 9: Do you have any thoughts on what we should be aiming for cost-wise and what timeframes we should aim for regarding feedback?

Faizi52996: i would say if we could save something by not setting up additional resource and not affect the output then it is cost effective. Timeframe would be as quick as possible once the changes are made.  I hope it answer your question.

WolfyUK:
It looks like you've put a lot of thought into this - nice one! 

I need to understand a bit more about the current dev workflow/pipelines to help make a decision around which option fits best. For example, if the plan is to run the different types of automated test concurrently (so as to increase efficiency of build runs) then you will definitely need some kind of separation between the pre-seeded test data to have confidence over the assertions (i.e. separate environments). If running these sequentially is preferred then the pre-/de-seed responsibilities of each type of test would be able to guarantee that the underlying data is in the expected format during each run. 

Also, what are the dependencies on the API vs. accessibility tests? Is it purely seeded DB data?

Faizi52996:
> It looks like you've put a lot of thought into this - nice one!
> 
> I need to understand a bit more about the current dev workflow/pipelines to help make a decision around which option fits best. For example, if the plan is to run the different types of automated test concurrently (so as to increase efficiency of build runs) then you will definitely need some kind of separation between the pre-seeded test data to have confidence over the assertions (i.e. separate environments). If running these sequentially is preferred then the pre-/de-seed responsibilities of each type of test would be able to guarantee that the underlying data is in the expected format during each run.
> 
> Also, what are the dependencies on the API vs. accessibility (and other UI-type) tests? Is it purely seeded DB data?

The tests could run concurrently but since there is no area(not that i can think of at this moment) in the application for users to update/delete data so even if they are running concurrently using the same data they don't essentially need to be in separate envs as reading/viewing data in different formats won't make cause any issues. 
The tests are not dependent on each other. They could share the DB if we go with pre seeded data.

spicy-rice (approved):

PR #80, opened 2024-01-17
integration tests added

J05h-L:
integration tests added to confirm behaviour of the schools, comparison, workforce, and investigation pages

### DIFF
--- a/web/tests/EducationBenchmarking.Web.Integration.Tests/WhenViewingSchoolComparison.cs
+++ b/web/tests/EducationBenchmarking.Web.Integration.Tests/WhenViewingSchoolComparison.cs
@@ -13,168 +13,92 @@ public class WhenViewingSchoolComparison : BenchmarkingWebAppClient
     {
     }
 
-    [Fact]
-    public async Task CanDisplayAcademy()
+    [Theory]
+    [InlineData(EstablishmentTypes.Academies)]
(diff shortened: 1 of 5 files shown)


web/tests/EducationBenchmarking.Web.Integration.Tests/WhenViewingSchoolComparison.cs
```diff
@@ -127,8 +164,14 @@ private static void AssertPageLayout(IHtmlDocument page, School school)
         var comparisonComponent = page.GetElementById("compare-your-school");
         Assert.NotNull(comparisonComponent);
         
-        var toolsHeadingSection = page.Body.SelectSingleNode("//main/div[4]");
-        DocumentAssert.Heading2(toolsHeadingSection, "More tools");
-        //TODO: Get and assert more tools section on page
+        var toolsListSection = page.Body.SelectSingleNode("//main/div[4]");
```
WolfyUK, line 167: Probably something for the future, but it might be safer to select elements by some kind of static identifier rather than an index.

WolfyUK: 👆🏼 This does seem to have been done in some places but not others

WolfyUK (approved):

PR #82, opened 2024-01-19
Fixed AzureRm provider to version 3.87

michael-fielding:
Fixed AzureRm provider to version 3.87 as a bug has been introduced in v3.88 that causes validation fails due to camel casing. There is an open bug report with terraform so this should be a temporary fix


WolfyUK:
Fix is being proposed in the next provider release as per https://github.com/hashicorp/terraform-provider-azurerm/issues/24560#issuecomment-1900197715

WolfyUK (approved):

PR #88, opened 2024-01-24
chore: ensures browser is closed after test scenarios

jrabbott:
Ensures browser is closed after test scenarios. 

PageHook is used to access current page, rather than  SpecFlow DI handle IPage instanace.

### DIFF
--- a/web/tests/EducationBenchmarking.Web.E2ETests/Pages/BenchmarkWorkforcePage.cs
+++ b/web/tests/EducationBenchmarking.Web.E2ETests/Pages/BenchmarkWorkforcePage.cs
@@ -1,4 +1,6 @@
 ﻿using EducationBenchmarking.Web.E2ETests.Helpers;
+using EducationBenchmarking.Web.E2ETests.Hooks;
+using EducationBenchmarking.Web.E2ETests.TestSupport;
 using Microsoft.Playwright;
 using Xunit;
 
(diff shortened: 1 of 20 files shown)


web/tests/EducationBenchmarking.Web.E2ETests/Pages/BenchmarkWorkforcePage.cs
```diff
@@ -57,7 +59,7 @@ public async Task AssertPage()
             await chart.ShouldBeVisible();
         }
 
-        var schoolWorkForceExpectedOptions = new[] { "total", "headcount per FTE", "pupils per staff role" };
+        var schoolWorkForceExpectedOptions = new[] { "total", "pupils per staff role" };
```
Faizi52996, line 62: the dropdown should have 'headcount per FTE' in it as per the requirement.

jrabbott: Recheck the user story - the "Headcount per FTE" option is not present.

For school workforce chart the value when this dimension is select will always be 1 (School workforce / school workforce).

Faizi52996: just double checked the story and current implemented system.  for school workforce FTE -> 'the percentage of workforce' option should not present in the dropdown as it would always be 1.
and for school workforce(headcount) -> 'head count per FTE' should not be present in the dropdown.

Faizi52996 (approved):

PR #101, opened 2024-01-25
Backlink - WiP

J05h-L:
first pass at adding Backlink to _Layout. Work in progress


web/src/EducationBenchmarking.Web/Views/Shared/_Layout.cshtml
```diff
@@ -82,8 +82,8 @@
         </p>
     </div>
 
-    <breadcrumb></breadcrumb>
-
+    <navigation></navigation>
```
WolfyUK, line 85: Please excuse my ignorance as I'm new to tag helpers, but how does this element link back to the custom `NavigationTagHelper`? Is it just resolved by naming convention?

jrabbott: `@addTagHelper *, EducationBenchmarking.Web` imports any custom helpers

Then convention is used. However I would prefer the explicit data attribute to be used. 

`[HtmlTargetElement("navigation", TagStructure = TagStructure.WithoutEndTag)] `

J05h-L: data attribute added

jrabbott (approved):

PR #105, opened 2024-01-26
Replace redundant resource type and switch to readytorun output

jrabbott:
Replace redundant resource type and switch to ReadyToRun output

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service_plan
https://learn.microsoft.com/en-us/dotnet/core/deploying/ready-to-run

### DIFF
--- a/web/terraform/main.tf
+++ b/web/terraform/main.tf
@@ -53,23 +53,20 @@ resource "azurerm_resource_group" "resource-group" {
   tags     = local.common-tags
 }
 
-resource "azurerm_app_service_plan" "education-benchmarking-asp" {
+resource "azurerm_service_plan" "education-benchmarking-asp" {
   name                = "${var.environment-prefix}-education-benchmarking-asp"
   location            = azurerm_resource_group.resource-group.location
   resource_group_name = azurerm_resource_group.resource-group.name
-
-  sku {
-    tier = var.sizing[var.environment].tier
-    size = var.sizing[var.environment].size
-  }
-  tags = local.common-tags
+  os_type             = "Windows"
+  sku_name            =  var.sizing[var.environment].size
+  tags                = local.common-tags
 }
 
 resource "azurerm_windows_web_app" "education-benchmarking-as" {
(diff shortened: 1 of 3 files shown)


web/terraform/main.tf
```diff
@@ -53,23 +53,20 @@ resource "azurerm_resource_group" "resource-group" {
   tags     = local.common-tags
 }
 
-resource "azurerm_app_service_plan" "education-benchmarking-asp" {
+resource "azurerm_service_plan" "education-benchmarking-asp" {
```
WolfyUK, line 56: Will destroying the old service plan affect the dependent app service?

WolfyUK:
I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?

web/terraform/main.tf
jrabbott: Yes, there will need to be some manual intervention as there is a "clean" migration path,

jrabbott:
> I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?



> I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?

I'm not sure the 2 are related.

WolfyUK:
> > I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?
> 
> I'm not sure the 2 are related.

@michael-fielding Was this the reason for pinning the provider version last week, or was it something else?

michael-fielding:
> I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?

michael-fielding:
> > > I guess once the app service plan is updated we can remove the pinned `azurerm` provider version number (see #82)?
> > 
> > 
> > I'm not sure the 2 are related.
> 
> @michael-fielding Was this the reason for pinning the provider version last week, or was it something else?

No, the reason for pinning the version was the bug introduced when Terraform started enforcing case-sensitivity in the "serverFarms" part of the app-service path

michael-fielding (approved):

PR #107, opened 2024-01-29
Corrects logging set in function apps

jrabbott:
Corrects logging to use app insights connection string (key is becoming redundant).

Adds structured logging using Serilog.

### DIFF
--- a/platform/terraform/main.tf
+++ b/platform/terraform/main.tf
@@ -90,15 +90,15 @@ resource "azurerm_key_vault_secret" "platform-search-key" {
 }
 
 module "benchmark-fa" {
-  source                   = "./modules/functions"
-  function-name            = "benchmark"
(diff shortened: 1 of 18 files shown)


platform/terraform/main.tf
WolfyUK, line 1: Do we run `terraform fmt` in the pipeline? The spacing change here might break it if so.

WolfyUK: (Also in the other `.tf` changes)

jrabbott: The spacing change won't break it. However I had considered adding to the pipeline, but a number of reason why I didn't - help to have a chat.

jrabbott: Will run fmt locally and push.... reads nicer

WolfyUK (approved):

PR #109, opened 2024-01-29
feat: Rider multi-project run config

WolfyUK:


platform/.run/EducationBenchmarking.Platform.Api.Benchmark.run.xml
```diff
@@ -0,0 +1,23 @@
...
+    <option name="PROJECT_EXE_PATH_TRACKING" value="1" />
+    <option name="PROJECT_ARGUMENTS_TRACKING" value="0" />
+    <option name="PROJECT_WORKING_DIRECTORY_TRACKING" value="1" />
+    <option name="PROJECT_KIND" value="RunnableProjectKind(name=AzureFunctions)" />
+    <option name="PROJECT_TFM" value="net6.0" />
+    <option name="FUNCTION_NAMES" value="" />
+    <browser url="http://localhost:7072" />
+    <method v="2">
+      <option name="BuildFunctionsProject" enabled="true" />
+      <option name="RunAzuriteTask" enabled="true" />
```
jrabbott, line 20: Might be an open question to the team about how they have Azurite setup? I tend to run Azurite in docker.

Oddly you don't need Azurite running to run func apps locally,

jrabbott (changes requested):

platform/.run/EducationBenchmarking.Platform.Api.Benchmark.run.xml
WolfyUK: Hmm. I didn't even notice that in there. Not quite sure when Rider would have even asked me to configure this - I probably just clicked a 'sure, OK' button!

jrabbott: Rider doesn't ask... it just does. It burnt me about 6 months when the Azurite I thought was being used, wasn't the one I thought!!

PR #113, opened 2024-01-30
chore: set ReadyToRun option on publish

jrabbott:
Set ReadyToRun option on publish and update proj settings to ensure package locks are up to date.

### DIFF
--- a/platform/src/abstractions/EducationBenchmarking.Platform.Functions/EducationBenchmarking.Platform.Functions.csproj
+++ b/platform/src/abstractions/EducationBenchmarking.Platform.Functions/EducationBenchmarking.Platform.Functions.csproj
@@ -4,6 +4,7 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
         <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+        <RuntimeIdentifiers>win-x86</RuntimeIdentifiers>
     </PropertyGroup>
     <ItemGroup>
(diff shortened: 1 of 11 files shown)


platform/src/abstractions/EducationBenchmarking.Platform.Functions/EducationBenchmarking.Platform.Functions.csproj
```diff
@@ -4,6 +4,7 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
         <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+        <RuntimeIdentifiers>win-x86</RuntimeIdentifiers>
```
WolfyUK, line 7: Will these additions affect only Azure-deployed functions, or is it likely to break debugging on the Mac, for example?

jrabbott: It's only just for publish so shouldn't break run/debug locally.

https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#runtimeidentifiers

jrabbott: linux-x64 has been setup on web solution and all is still working fine for folks locally.

WolfyUK (approved):

PR #120, opened 2024-01-31
Fix Accordion A11y issues

J05h-L:


jrabbott (changes requested):
Need to fix E2E tests, as changing the element ID has broken those.

jrabbott (approved):

PR #130, opened 2024-02-02
Refactor run terraform so they can be reused

jrabbott:
Refactor run terraform so they can be reused


WolfyUK (approved):
I don't yet have `terraform` installed locally to validate, but this looks good. I guess there will be follow-up PRs for the other pipelines to share the same steps?

PR #132, opened 2024-02-05
fix: workforce headcount per FTE

J05h-L:
- corrected calculation for HeadcountPerFTE in CalculateWorkforceValue. now this is workforce headcount / specific FTE (value)

- removed dropdown value on chart/table for School workforce (Headcount). This is now the only trivial result for the data.

### DIFF
--- a/front-end-components/src/views/compare-your-workforce/partials/headcount.tsx
+++ b/front-end-components/src/views/compare-your-workforce/partials/headcount.tsx
@@ -8,6 +8,7 @@ import {
   PupilsPerStaffRole,
   WorkforceCategories,
   HeadcountPerFTE,
+  PercentageOfWorkforce,
 } from "src/components";
 import { ChartDimensionContext } from "src/contexts";
 import { HeadcountProps } from "src/views/compare-your-workforce/partials";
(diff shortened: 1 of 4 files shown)


front-end-components/src/views/compare-your-workforce/partials/headcount.tsx
```diff
@@ -56,7 +57,9 @@ export const Headcount: React.FC<HeadcountProps> = (props) => {
       >
         <h2 className="govuk-heading-m">School workforce (Headcount)</h2>
         <ChartDimensions
-          dimensions={WorkforceCategories}
+          dimensions={WorkforceCategories.filter(
+            (category) => category !== HeadcountPerFTE
```
jrabbott, line 61: Need to remove "PercentageOfWorkforce" from the selection too.

jrabbott (changes requested):

front-end-components/src/views/compare-your-workforce/partials/headcount.tsx
J05h-L: pushed with this removed

jrabbott (approved):

PR #138, opened 2024-02-06
chore: resolve warnings in platform solution

jrabbott:
Before submitting this PR as ready for review, please make sure:

- [x] Work items have been linked
- [x] Your branch has been rebased onto master

***********

Resolves all warnings in platform solution

_**no work item in backlog**_

### DIFF
--- a/platform/src/abstractions/EducationBenchmarking.Platform.Domain/DataObjects/SchoolRatingDataObject.cs
+++ b/platform/src/abstractions/EducationBenchmarking.Platform.Domain/DataObjects/SchoolRatingDataObject.cs
@@ -1,18 +1,19 @@
 using System.Diagnostics.CodeAnalysis;
+using Newtonsoft.Json;
 
 namespace EducationBenchmarking.Platform.Domain.DataObjects;
 
 [ExcludeFromCodeCoverage]
-public class SchoolRatingDataObject
+public record SchoolRatingDataObject
 {
-    public string Term { get; set; }
-    public string OverallPhase { get; set; }
-    public string AssessmentArea { get; set; }
-    public string Divisor { get; set; }
-    public string Size { get; set; }
-    public string FSM { get; set; }
+    public string? Term { get; set; }
+    public string? OverallPhase { get; set; }
+    public string? AssessmentArea { get; set; }
+    public string? Divisor { get; set; }
+    public string? Size { get; set; }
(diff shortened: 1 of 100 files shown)


platform/src/abstractions/EducationBenchmarking.Platform.Domain/DataObjects/SchoolRatingDataObject.cs
```diff
@@ -1,18 +1,19 @@
...
-    public string AssessmentArea { get; set; }
-    public string Divisor { get; set; }
-    public string Size { get; set; }
-    public string FSM { get; set; }
+    public string? Term { get; set; }
+    public string? OverallPhase { get; set; }
+    public string? AssessmentArea { get; set; }
+    public string? Divisor { get; set; }
+    public string? Size { get; set; }
+    [JsonProperty("FSM")] public string? Fsm { get; set; }
```
WolfyUK, line 14: Some of the other objects in this folder use constants for these attributes. Is it worth rolling this out everywhere?

jrabbott: I'm hoping we won't need them when we produce our own data pipeline, and we'll follow standard JSON variable naming conventions.

platform/src/apis/EducationBenchmarking.Platform.Api.Establishment/LocalAuthoritiesFunctions.cs
```diff
@@ -12,31 +12,31 @@ namespace EducationBenchmarking.Platform.Api.Establishment;
 public class LocalAuthoritiesFunctions
 {
     
-    [FunctionName(nameof(GetLocalAuthorityAsync))]
```
WolfyUK, line 15: I'm not a huge fan of this ReSharper recommendation to drop `Get` but I'll live with it elsewhere. On an `HttpTrigger` I really think it would be handy to keep them though to help distinguish between the HTTP method definition.

jrabbott: I was surprised to see the warning. With the other `HttpTrigger` they are pre-fixed with the operation (e.g. query, search, create).

I agree it would be handy to have something to help distinguish - how about "SingleLocalAuthorityAsync"?

WolfyUK (approved):

PR #170, opened 2024-02-13
Added `ITestOutputHelper` to A11y tests

WolfyUK:
### Context
HTTP errors due to misconfigured A11y tests were being consumed by the application making debugging difficult.

Output HTTP responses to the `ITestOutputHelper` to give better visibility over what the A11y tests are doing.

### Guidance to review 
Running the A11y tests locally should output all HTTP responses. 

### Checklist
- ~~[ ] Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)~~

### DIFF
--- a/web/tests/EducationBenchmarking.Web.A11yTests/PageBase.cs
+++ b/web/tests/EducationBenchmarking.Web.A11yTests/PageBase.cs
@@ -1,6 +1,3 @@
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
 using Deque.AxeCore.Commons;
 using Deque.AxeCore.Playwright;
 using Microsoft.Playwright;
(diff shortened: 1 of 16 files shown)


Chris-Lorch (approved):

jrabbott:
This doesn't feel right. If we're wanting to set the output helper in the WebDrover then I'd suggest not set it as a IClassFixture.

Instead I would pass the output helper explicitly in the constructor for the WebDriver. Then follow constructor/disposal pattern for the test class.

https://xunit.net/docs/shared-context

WolfyUK:
> This doesn't feel right. If we're wanting to set the output helper in the WebDrover then I'd suggest not set it as a IClassFixture.
> 
> Instead I would pass the output helper explicitly in the constructor for the WebDriver. Then follow constructor/disposal pattern for the test class.
> 
> https://xunit.net/docs/shared-context

Doing in the `IClassFixture` doesn't work - it seems as though this is the only pattern that does:
- https://stackoverflow.com/a/59222453/504477
- https://github.com/xunit/visualstudio.xunit/issues/214

WolfyUK:
> We shouldn't be use IClassFixture for WebDriver if we need to set dependencies.

Doing in the `IClassFixture` doesn't work - it seems as though this is the only pattern that does:
- https://stackoverflow.com/a/59222453/504477
- https://github.com/xunit/visualstudio.xunit/issues/214

jrabbott (approved):
We could probably clean up the page assessment, but something for a future change.

PR #270, opened 2024-02-20
Addition of a generic Line Chart component using Recharts

WolfyUK:
### Context
[AB#196192](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/196192)

Addition of a generic Line Chart component using Recharts. Addition of `<Stat>` component for highlighting specific key/value pair. Refactor of shared chart props.

### Guidance to review 
Run `front-end-components` to see the new chart in action

### Checklist

### DIFF
--- a/front-end-components/src/components/charts/stat/component.tsx
+++ b/front-end-components/src/components/charts/stat/component.tsx
@@ -27,6 +27,7 @@ export function Stat<TData extends ChartDataSeries>(props: StatProps<TData>) {
     return { label: entry[1][seriesLabelField], value: entry[1][valueField] };
   }, [data, displayIndex, seriesLabelField, valueField]);
 
+  // do not render anything if a match could not be located based on the available data
   if (!entry) {
     return null;
   }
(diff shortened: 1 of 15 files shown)


front-end-components/src/components/charts/stat/component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const entry = useMemo(() => {
+    const entry = Object.entries(data)[displayIndex];
+    if (!entry) {
+      return null;
+    }
+
+    return { label: entry[1][seriesLabelField], value: entry[1][valueField] };
+  }, [data, displayIndex, seriesLabelField, valueField]);
+
+  if (!entry) {
```
jrabbott, line 30: This throw me for a moment - thinking is was rogue. Would it be better made this as a ternary on the main return?

WolfyUK: Could do, but in React I find it much more readable just to bomb out as soon as you can with `null` if there is nothing left to render

WolfyUK: Happy to add in a comment instead?

jrabbott: Yeah, comment is fine.

jrabbott (approved):

PR #416, opened 2024-03-13
Adds SQL Server & Database

michael-fielding:
Creates SQL server and database for comparator sets using Terraform

### Checklist
- ~~[ ] Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)~~

### DIFF
--- a/platform/terraform/main.tf
+++ b/platform/terraform/main.tf
@@ -84,6 +84,45 @@ resource "azurerm_cosmosdb_sql_database" "cosmosdb-container" {
   resource_group_name = azurerm_resource_group.resource-group.name
 }
 
+resource "random_password" "sql_admin_password" {
+  length = 16
+  special = true
+}
+
+resource "azurerm_key_vault_secret" "platform-sql-admin-username" {
+  name         = "ebis-sql-admin-username"
+  value        = "${var.environment-prefix}-ebis-sql-admin"
+  key_vault_id = data.azurerm_key_vault.key-vault.id
+}
+
+resource "azurerm_key_vault_secret" "platform-sql-admin-password" {
+  name         = "ebis-sql-admin-password"
+  value        = random_password.sql_admin_password.result
+  key_vault_id = data.azurerm_key_vault.key-vault.id
+}
+
+resource "azurerm_mssql_server" "sql_server" {
+  name                         = "${var.environment-prefix}-ebis-sql"
+  version                      = "12.0"
+  resource_group_name          = azurerm_resource_group.resource-group.name
+  location                     = azurerm_resource_group.resource-group.location
+  administrator_login          = "${var.environment-prefix}-ebis-sql-admin"
+  administrator_login_password = random_password.sql_admin_password.result
+  tags                         = local.common-tags
+}
+
+resource "azurerm_mssql_database" "sql_db" {
+  name      = "ebis-data"
+  server_id = azurerm_mssql_server.sql_server.id
+  tags      = local.common-tags
+}
+
+resource "azurerm_key_vault_secret" "platform-sql-connection-string" {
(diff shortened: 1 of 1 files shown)


platform/terraform/main.tf
```diff
@@ -84,6 +84,45 @@ resource "azurerm_cosmosdb_sql_database" "cosmosdb-container" {
...
 }
 
+resource "random_password" "sql_admin_password" {
+  length = 16
+  special = true
+}
+
+resource "azurerm_key_vault_secret" "platform-sql-admin-username" {
+  name         = "ebis-sql-admin-username"
+  value        = "${var.environment-prefix}-ebis-sql-admin"
```
jrabbott, line 94: This feels over kill? do we need env prefix on the user name?

platform/terraform/main.tf
```diff
@@ -84,6 +84,45 @@ resource "azurerm_cosmosdb_sql_database" "cosmosdb-container" {
...
+  tags                         = local.common-tags
+}
+
+resource "azurerm_mssql_database" "sql_db" {
+  name      = "ebis-data"
+  server_id = azurerm_mssql_server.sql_server.id
+  tags      = local.common-tags
+}
+
+resource "azurerm_key_vault_secret" "platform-sql-connection-string" {
```
jrabbott, line 120: Not sure we need the full connection string in key vault?

platform/terraform/main.tf
```diff
@@ -84,6 +84,45 @@ resource "azurerm_cosmosdb_sql_database" "cosmosdb-container" {
   resource_group_name = azurerm_resource_group.resource-group.name
 }
 
+resource "random_password" "sql_admin_password" {
```
jrabbott, line 87: I'm not sure this meets the password policy for SQL

https://learn.microsoft.com/en-us/sql/relational-databases/security/password-policy?view=sql-server-ver16&redirectedfrom=MSDN

platform/terraform/main.tf
```diff
@@ -84,6 +84,45 @@ resource "azurerm_cosmosdb_sql_database" "cosmosdb-container" {
...
+  name                         = "${var.environment-prefix}-ebis-sql"
+  version                      = "12.0"
+  resource_group_name          = azurerm_resource_group.resource-group.name
+  location                     = azurerm_resource_group.resource-group.location
+  administrator_login          = "${var.environment-prefix}-ebis-sql-admin"
+  administrator_login_password = random_password.sql_admin_password.result
+  tags                         = local.common-tags
+}
+
+resource "azurerm_mssql_database" "sql_db" {
```
jrabbott, line 114: Should we not proactively be setting the SKU?

jrabbott:
AB#199507 (AB#200089)

platform/terraform/main.tf
jrabbott: Think we might need to set "override_special" with a given set of chars

michael-fielding: What characters do you suggest? MSDN doesn't provide a definitive list

jrabbott: I'd suggest `!@#$*()-_=+` to avoid escape issues.

platform/terraform/main.tf
michael-fielding: Will remove it

platform/terraform/main.tf
michael-fielding: I'll take it out

platform/terraform/main.tf
jrabbott: Still showing?

jrabbott (approved):

PR #469, opened 2024-03-19
Setting and Docs tidy up for Web frontend.

colinbull:
### Context
This is a minor PR that adds the following 

* .gitgnore for local dev SQL scenarios
* An extra note about `dotnet user-secrets init` in README
* global.json added to ensure that the .NET version is contolled by the solution
* removed DfE settings.

### DIFF
--- a/web/src/Web.App/appsettings.Development.json
+++ b/web/src/Web.App/appsettings.Development.json
@@ -21,18 +21,6 @@
     "ContainerName": "sessions",
     "DatabaseName": "session-data"
   },
-  "DFESignInSettings": {
-    "APISecret": "some-secret",
-    "APIUri": "https://pp-api.signin.education.gov.uk",
-    "Audience": "signin.education.gov.uk",
-    "CallbackPath": "/auth/cb",
-    "ClientID": "some-client-id",
-    "ClientSecret": "some-client-id",
-    "Issuer": "some-issuer",
-    "MetadataAddress": "https://pp-oidc.signin.education.gov.uk/.well-known/openid-configuration",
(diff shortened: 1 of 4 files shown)


web/src/Web.App/appsettings.Development.json
jrabbott, line 1: Probably ought to update the readme to include the expected DSI setting keys for the secrets.

web/src/Web.App/appsettings.Development.json
```diff
@@ -21,18 +21,6 @@
     "ContainerName": "sessions",
     "DatabaseName": "session-data"
   },
-  "DFESignInSettings": {
```
WolfyUK, line 24: Personally I like keeping these here, even though they get overridden by dotnet user secrets. It lets the Devs know that there is some additional configuration required, and the base values can be copy-and-pasted directly into the relevant `secrets.json` file if required, rather than using the CLI.

colinbull: I have added this to the README.md now. My worry is keeping those settings and people updating the settings and not realising that it is actually the user-secrets they need to update, when they have both set, it is easy to forget. At least if they don't have these settings in the first place it would just fail.

While I don't have a strong opinion if this is the approach we are going to take having the secrets templated both in here and in the secrets. We should then make sure we do it for all settings.

jrabbott: @colinbull I'll pick up updating the readme with templated secrets files on a separate PR.

jrabbott (approved):

PR #660, opened 2024-04-28
chore: adds school and trust authorization attributes

jrabbott:
### Context
Authorization required CFP and budget and forecasting - whereby school and trusts can only access there own data.

Adds two authorization attributes to validate user claims against route values

### Guidance to review 
Claims to be set as part of DSI integration

### Checklist
- [x] ~~Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)~~

### DIFF
--- a/web/src/Web.App/Attributes/TrustAuthorizationAttribute.cs
+++ b/web/src/Web.App/Attributes/TrustAuthorizationAttribute.cs
@@ -9,9 +9,9 @@ public class TrustAuthorizationAttribute : AuthorizeAttribute, IAuthorizationFil
 {
     public void OnAuthorization(AuthorizationFilterContext context)
     {
-        var urn = context.RouteData.Values["companyNumber"]?.ToString();
+        var companyNumber = context.RouteData.Values["companyNumber"]?.ToString();
 
-        var isValid = context.HttpContext.User.Claims.Any(c => urn != null && c.Type == "Trusts" && c.Value.Contains(urn));
+        var isValid = context.HttpContext.User.Claims.Any(c => companyNumber != null && c.Type == "Trusts" && c.Value.Contains(companyNumber));
         if (!isValid)
         {
             context.Result = new UnauthorizedResult();


web/src/Web.App/Attributes/SchoolAuthorizationAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Web.App.Attributes;
+
+[AttributeUsage(AttributeTargets.Class)]
+public class SchoolAuthorizationAttribute : AuthorizeAttribute, IAuthorizationFilter
+{
+    public void OnAuthorization(AuthorizationFilterContext context)
+    {
+        var urn = context.RouteData.Values["urn"]?.ToString();
+
+        var isValid = context.HttpContext.User.Claims.Any(c => urn != null && c.Type == "Schools" && c.Value.Contains(urn));
```
WolfyUK, line 14: Do we need null checks on `User.Claims` always or `Claim.Value` or is this safe enough as-is?

web/src/Web.App/Attributes/TrustAuthorizationAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Web.App.Attributes;
+
+[AttributeUsage(AttributeTargets.Class)]
+public class TrustAuthorizationAttribute : AuthorizeAttribute, IAuthorizationFilter
+{
+    public void OnAuthorization(AuthorizationFilterContext context)
+    {
+        var companyNumber = context.RouteData.Values["companyNumber"]?.ToString();
+
+        var isValid = context.HttpContext.User.Claims.Any(c => companyNumber != null && c.Type == "Trusts" && c.Value.Contains(companyNumber));
```
WolfyUK, line 14: Do we need null checks on User.Claims or Claim.Value or is this safe enough as-is?

web/src/Web.App/Attributes/SchoolAuthorizationAttribute.cs
jrabbott: Should be enough as-is. But we can ensure we have integration tests to cover this when we fully integrate DSI.

web/src/Web.App/Attributes/TrustAuthorizationAttribute.cs
jrabbott: Should be enough as-is. But we can ensure we have integration tests to cover this when we fully integrate DSI.

WolfyUK (approved):

PR #675, opened 2024-04-30
increased the default waiting time to 30secs

Faizi52996:
### Context
there are some flaky tests because of timeout. increased it to 30 secs now

n/a
### Guidance to review 
n/a
### Checklist
~~- [ ] Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)~~
~~- [ ] You have run all unit/integration tests and they pass~~

### DIFF
--- a/web/tests/Web.E2ETests/Extensions/LocatorExtensions.cs
+++ b/web/tests/Web.E2ETests/Extensions/LocatorExtensions.cs
@@ -21,7 +21,7 @@ public static class LocatorExtensions
     }
     public static async Task<ILocator> ShouldBeVisible(this ILocator locator)
     {
-        await Assertions.Expect(locator).ToBeVisibleAsync();
+        await Assertions.Expect(locator).ToBeVisibleAsync(new LocatorAssertionsToBeVisibleOptions { Timeout = 30000 });
 
         return locator;
     }


jrabbott (commented):
If pages/elements are taking longer than 30 secs, I think we need to review it.

Faizi52996:
> If pages/elements are taking longer than 30 secs, I think we need to review it.

currently the default timeout is 5s but in our case sometime the chart takes more time to load causing flakiness. if they would take more than 30s(worst case) then definitely need to review them.

jrabbott:
@Faizi52996 why are we only just seeing this?

Faizi52996:
@jrabbott, the recent history pages changes seem to have affected more with this. Previously we made the use of ` await page.WaitForLoadStateAsync(LoadState.NetworkIdle);` in few places but now i think it would make sense to increase the waiting time rather than waiting for loadstate at different stages.

Faizi52996:
@jrabbott , I see what you mean. This probably is not the root cause in this case. closing this PR for now for further troubleshooting.

PR #752, opened 2024-05-14
Custom Data 'Totals' calculations and additional validation rules

WolfyUK:
### Context
[AB#209239](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/209239) [AB#198080](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/198080)

Additional validation rules on Workforce page. Client side calculation of Income/Expenditure totals on Financial Data page.

### Guidance to review 
Vanilla JS is used for the client side calculations here within the Web project, so there is no need to copy over `front-end-components` to test locally. There is a case for refactoring this script into something that could get `gulp`ed from TypeScript in `AssetSrc` into `wwwroot` and then importing by filename in the view rather than polluting the latter with lots of JS. Happy to do so if the reviewer insists!

### Checklist

### DIFF
--- a/web/src/Web.App/Views/SchoolCustomDataChange/FinancialData.cshtml
+++ b/web/src/Web.App/Views/SchoolCustomDataChange/FinancialData.cshtml
@@ -6,7 +6,7 @@
 @await Component.InvokeAsync("EstablishmentHeading", new { title = PageTitles.SchoolChangeDataFinancialData, name = Model.School.Name, id = Model.School.Urn, kind = OrganisationTypes.School })
 
 @await Html.PartialAsync("_ErrorSummary")
-@using (Html.BeginForm("FinancialData", "SchoolCustomDataChange", new { urn = Model.School.Urn }, FormMethod.Post, true, new { novalidate = "novalidate" }))
+@using (Html.BeginForm("FinancialData", "SchoolCustomDataChange", new { urn = Model.School.Urn }, FormMethod.Post, true, new { novalidate = "novalidate", id = "form-custom-data-financial" }))
 {
     <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-financial-data">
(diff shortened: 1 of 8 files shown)


web/src/Web.App/Views/SchoolCustomDataChange/FinancialData.cshtml
jrabbott, line 1: I'm not sure the front-end can actually calculate the new totals? I think this will only be known once the data pipeline has run with the custom values.

WolfyUK: I queried this with Assad which is why we went with this implementation. Yes, there are other financials outside of what is available here but the difference between the 'current' and 'custom' can be used to determine the 'custom' totals based on the base figures. If you are not convinced then perhaps we can chat about this with Assad?

jrabbott: I'm not convinced that it should be as we're duplicating calculation logic - let's leave it for now, we can remove/revisit later on once it's fully integrated with the pipeline and we have an e2e.

I'm not even sure why we have the totals section - Revenue reserve value won't change, and there is only one custom income value.

It would be good to gulp-ify the JS.

WolfyUK: Would you like it kept as-is, or `gulp`ed, even if it might be nuked at some point?

jrabbott: Good shout - let's refactor later

jrabbott (approved):

PR #759, opened 2024-05-16
chore: adds core database and initial scripts

jrabbott:
### Context

1. Adds core database infrastructure
2. Solution for dbUp program 
3. Initial table scripts for Comparator sets, User defined comparator sets, Metric RAG and Local authority
4. Update pipeline to build and deploy/execute db schema

### DIFF
--- a/core-infrastructure/src/db/Core.Database/Scripts/001-CreateComparatorSetTable.sql
+++ b/core-infrastructure/src/db/Core.Database/Scripts/001-CreateComparatorSetTable.sql
@@ -1,0 +1,16 @@
+IF NOT EXISTS(SELECT *
+               FROM INFORMATION_SCHEMA.TABLES
+               WHERE table_name = 'ComparatorSet')
+    BEGIN
+        CREATE TABLE dbo.ComparatorSet
+        (
+            RunType  nvarchar(50)  NOT NULL,
+            RunId    nvarchar(50)  NOT NULL,
+            UKPRN    nvarchar(50)  NOT NULL,
+            SetType  nvarchar(50)  NOT NULL,
+            Pupil    nvarchar(max) NULL,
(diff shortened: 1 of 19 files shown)


core-infrastructure/terraform/databases.tf
```diff
@@ -0,0 +1,233 @@
...
+
+  network_rules {
+    default_action = "Deny"
+    bypass = [
+      "AzureServices"
+    ]
+  }
+}
+
+resource "azurerm_storage_account" "sql-vulnerability-storage" {
```
colinbull, line 196: Can't the audit storage and vunerability storage all go to one core storage account for all databases

jrabbott: Just checked container names on threat and audit. All three could go into a single storage account.

colinbull (approved):

core-infrastructure/src/db/Core.Database/Scripts/001-CreateComparatorSetTable.sql
```diff
@@ -0,0 +1,16 @@
+IF NOT EXISTS(SELECT *
+               FROM INFORMATION_SCHEMA.TABLES
+               WHERE table_name = 'ComparatorSet')
+    BEGIN
+        CREATE TABLE dbo.ComparatorSet
+        (
+            RunType  nvarchar(50)  NOT NULL,
```
PsypherPunk, line 7: More of a general question about the columns: a lot are `nvarchar(50)` with only a few exceptions, is this because we don't know the range of values that _should_ be in there and can't constrain things further?

jrabbott: @PsypherPunk exactly that. Didn't want to constrain things too much at this point.

PR #888, opened 2024-06-03
feat: adds comparator list pages

jrabbott:
### Context
AB#210705

Adds school comparator list pages

### Guidance to review 
N/A

### Checklist

### DIFF
--- a/web/src/Web.App/Domain/ComparatorSet.cs
+++ b/web/src/Web.App/Domain/ComparatorSet.cs
@@ -12,5 +12,5 @@ public record ComparatorSet
 
 public record ComparatorSetUserDefined
 {
-
+    public string[] Set { get; set; } = Array.Empty<string>();
 }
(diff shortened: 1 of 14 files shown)


web/src/Web.App/Domain/ComparatorSet.cs
```diff
@@ -12,5 +12,5 @@ public record ComparatorSet
 
 public record ComparatorSetUserDefined
 {
-
+    public string[] Set { get; set; } = Array.Empty<string>();
```
WolfyUK, line 15: Is `Set` too generic a name? Are these only ever going to be URNs?

web/src/Web.App/Views/SchoolComparators/Index.cshtml
```diff
@@ -53,12 +54,48 @@
                     <li>pupils eligible for free school meals (FSM)</li>
                     <li>pupils with special educational needs (SEN)</li>
                 </ul>
-                <h3 class="govuk-heading-s">Schools we've chosen</h3>
+                <table class="govuk-table">
```
WolfyUK, line 57: This table looks very familiar! I might refactor into a partial as part of my current work and include an optional extra column where I need to support the 'removal' from a user defined set

web/src/Web.App/Domain/ComparatorSet.cs
jrabbott: Yeah, it's just a list of URN's

jrabbott: Not keep on changing as it's mapped all the way to the db column

PR #892, opened 2024-06-04
Submit pending user defined comparator set and poll on result

WolfyUK:
### Context
[AB#212120](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/212120) [AB#210705](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/210705)

Submit pending user defined comparator set to data pipeline and poll until corresponding user data entry completes or fails.

### Guidance to review 
Running locally the user data entry never moves out of the `pending` state, so this will need to be done manually to complete testing of the journey end-to-end. e.g.:

```sql
update [UserData] set [Status] = 'complete' where id = '<guid>'
```

### Checklist

### DIFF
--- a/web/src/Web.App/Controllers/SchoolComparatorsCreateByController.cs
+++ b/web/src/Web.App/Controllers/SchoolComparatorsCreateByController.cs
@@ -4,7 +4,6 @@ using Microsoft.FeatureManagement.Mvc;
 using Web.App.Attributes;
 using Web.App.Domain;
 using Web.App.Extensions;
-using Web.App.Identity;
 using Web.App.Infrastructure.Apis;
 using Web.App.Infrastructure.Extensions;
 using Web.App.Services;
@@ -13,7 +12,7 @@ using Web.App.ViewModels;
 namespace Web.App.Controllers;
 
 [Controller]
-//todo: reinstate once DSI permissions resolved [SchoolAuthorization]
+[SchoolAuthorization]
 [FeatureGate(FeatureFlags.UserDefinedComparators)]
 [Route("school/{urn}/comparators/create/by")]
 public class SchoolComparatorsCreateByController(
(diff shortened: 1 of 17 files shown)


web/src/Web.App/Controllers/SchoolComparatorsCreateByController.cs
```diff
@@ -11,14 +12,15 @@
 namespace Web.App.Controllers;
 
 [Controller]
-//todo: reinstate once DSI permissions resolved [SchoolAuthorization]
+[SchoolAuthorization]
```
jrabbott, line 15: Doesn't required school authorisation, just authorisation.

An authenticated user can create a user defined comparator set for any school, not just there own.

web/src/Web.App/Views/School/Index.cshtml
```diff
@@ -23,6 +23,34 @@
...
+                    <h3 class="govuk-notification-banner__heading">@(success ? "You are now benchmarking against your own set of schools" : "Unable to benchmark against your own set of schools")</h3>
+                    @if (!success)
+                    {
+                        <p class="govuk-body">
+                            Please try again later.
+                        </p>
+                    }
+                    <p>
+                        <a class="govuk-notification-banner__link"
+                           href="@Url.Action("Index", "SchoolComparatorsCreate", new { urn = Model.Urn })">
```
jrabbott, line 46: This can link to the UserDefined action in the SchoolComparators controller.

web/src/Web.App/Controllers/SchoolController.cs
```diff
@@ -18,9 +17,14 @@ public class SchoolController(
     : Controller
 {
     [HttpGet]
-    public async Task<IActionResult> Index(string urn)
+    public async Task<IActionResult> Index(
+        string urn,
+        [FromQuery(Name = "comparator-generated")] bool? comparatorGenerated)
```
jrabbott, line 22: Possible future improvement - use TempData rather than query param.

jrabbott (approved):

jrabbott:
@WolfyUK Feel free to resolve the comments and pick up on the next pass.

web/src/Web.App/Controllers/SchoolController.cs
WolfyUK: I think I'll leave this one as-is for the moment, as it will require change to the polling which is handled client side.

jrabbott (approved):

PR #898, opened 2024-06-05
Data pipeline/bfr logic2

Chris-Lorch:
### Context
Added in DNS logic fix for pre-processing pipeline.
Added in historic data consideration to BFR logic


data-pipeline/src/pipeline/main.py
```diff
@@ -248,33 +248,38 @@ def pre_process_bfr(set_type, year):
         raw_container, f"{set_type}/{year}/BFR_3Y_raw.csv", encoding="unicode-escape"
     )
 
-    bfr_sofa, bfr_3y, bfr, bfr_metrics = build_bfr_data(bfr_sofa, bfr_3y)
+    academies_y2 = prepare_data(
+        pd.read_parquet(
+            get_blob("pre-processed", f"{set_type}/{year-2}/academies.parquet")
```
PsypherPunk, line 253: Looking at the example message sent earlier:

```json
{
  "jobId": "5eabc249-e0e7-412c-9471-ebd05a66d4c4",
  "type": "comparator-set",
  "runType": "default",
  "runId": "9dbc9556-184b-45e6-bf85-3b3688807a9c",
  "year": "2022",
```

…`year` is a `str` when passed in, so the `year-2` calculation will fail (unless I'm missing an `int()` call somewhere.

PsypherPunk (commented):
One query on the use of `year` but broadly, I think this makes sense (insofar as I understand things).

data-pipeline/src/pipeline/main.py
Chris-Lorch: Good shout, I've added in some `int()`'s to deal with this

PsypherPunk: For the sake of clarity/simplicity, you can probably just add a:

```python
year = int(msg_payload["year"])
```

…line in `handle_msg()` and pass that `year` around, rather than passing `msg_payload["year"]` as we're currently doing.

jrabbott: @PsypherPunk if it's helpful I can change the datatype on the message?

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
```diff
@@ -0,0 +1,13 @@
+IF NOT EXISTS(SELECT *
+              FROM INFORMATION_SCHEMA.TABLES
+              WHERE table_name = 'BFR')
+    BEGIN
+        CREATE TABLE dbo.ComparatorSet
+        (
+            RunType         nvarchar(50)  NOT NULL,
+            RunId           nvarchar(50)  NOT NULL,
+            TrustUPIN       nvarchar(6)   NOT NULL,
```
jrabbott, line 9: Is there a reason why this is using TrustUPIN rather than CompanyNumber?

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
```diff
@@ -0,0 +1,13 @@
+IF NOT EXISTS(SELECT *
+              FROM INFORMATION_SCHEMA.TABLES
+              WHERE table_name = 'BFR')
+    BEGIN
+        CREATE TABLE dbo.ComparatorSet
```
jrabbott, line 5: Table name is incorrect

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
```diff
@@ -0,0 +1,13 @@
...
+              FROM INFORMATION_SCHEMA.TABLES
+              WHERE table_name = 'BFR')
+    BEGIN
+        CREATE TABLE dbo.ComparatorSet
+        (
+            RunType         nvarchar(50)  NOT NULL,
+            RunId           nvarchar(50)  NOT NULL,
+            TrustUPIN       nvarchar(6)   NOT NULL,
+
+            CONSTRAINT PK_ComparatorSet PRIMARY KEY (RunType, RunId, TrustUPIN)
```
jrabbott, line 11: Index name is incorrect

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
Chris-Lorch: It should be, I'll change that now.

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
Chris-Lorch: It should be, I'll change that now

data-pipeline/src/pipeline/main.py
PsypherPunk: TBF, that would probably be a more sensible change (and allow us to validate the incoming message a little more).

@Chris-Lorch, you can ignore me an assume it's an `int`, if that's the case.

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
jrabbott: We'll need to ensure that is added to the Trust table, otherwise we won't be able to link/lookup that data.

data-pipeline/src/pipeline/main.py
jrabbott: @Chris-Lorch / @PsypherPunk I'll make that change to the message

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
jrabbott: Thinking about it BudgetForecastReturn might be a more meaningful name.

core-infrastructure/src/db/Core.Database/Scripts/017-CreateBFRTable.sql
jrabbott: It will need implementing, as part of the BFR work.

data-pipeline/src/pipeline/main.py
Chris-Lorch: I'll leave it as is for now then if the data type in the message is specified

data-pipeline/src/pipeline/main.py
```diff
@@ -248,33 +248,38 @@ def pre_process_bfr(set_type, year):
         raw_container, f"{set_type}/{year}/BFR_3Y_raw.csv", encoding="unicode-escape"
     )
 
-    bfr_sofa, bfr_3y, bfr, bfr_metrics = build_bfr_data(bfr_sofa, bfr_3y)
+    academies_y2 = prepare_data(
+        pd.read_parquet(
+            get_blob("pre-processed", f"{set_type}/{int(year)-2}/academies.parquet")
```
PsypherPunk, line 253: We've agreed (as per a comment elsewhere) that `year` will be passed in as an `int` upstream so this should be safe now.

PsypherPunk (approved):
Just that one comment on the `int(year)` thing but LGTM.

PsypherPunk (approved):

PR #902, opened 2024-06-06
Comparator set result submission and results preview

WolfyUK:
### Context
[AB#212126](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/212126) [AB#212125](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/212125) [AB#210705](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/210705)

Once characteristics selected, post them to back end API to generate a comparator set. Preview the comparator set characteristics and then allow new user defined set to be submitted for updated benchmark data (as per existing flow).

### Guidance to review 
This should allow the user defined characteristic journey to be completely run through end-to-end.

### Checklist


web/src/Web.App/AssetSrc/scss/main.scss
jrabbott, line 1: We should avoid styling buttons as links, unless there is a user need for it. They aren't very accessible.

GOV.UK design system doesn't have tertiary buttons I suspect for a reason.

WolfyUK: I thought you might pick up on this. Happy to change it back to a standard button, but this differs from the prototype. Alternatively I could use query string instead (which was actually my first attempt), but this might look a little messy. I'd rather avoid session state.

jrabbott: Let's keep it as a button and styled as a secondary button.

WolfyUK: I've removed the `<button>` instead and reinstated use of the query string in order to match the prototype.

jrabbott (approved):

PR #916, opened 2024-06-10
feat: user-defined comparator sets

PsypherPunk:
relates #212505

[AB#212505](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/212505)

### Context

Changes required to accommodate user-defined comparator sets.

- initial check in `main.py` as to whether the incoming message holds a user-defined comparator set;
- if so, skip pre-processing;
- run `run_user_defined_rag()` (a variant of the existing `run_compute_rag()` and `compute_rag_for()` functions);
- run `compute_user_defined_rag()` (similar to existing `compute_rag()`, less the comparator-set derivation);
- write to blob-storage;
- write to database.

### Guidance to review 

- have I correctly understood the pipeline flow?
- are there any specific requirements around the blob-storage/database keys?

### Checklist

### DIFF
--- a/data-pipeline/src/pipeline/main.py
+++ b/data-pipeline/src/pipeline/main.py
@@ -6,6 +6,7 @@ from contextlib import suppress
 import pandas as pd
 import tornado.iostream
 from azure.core.exceptions import ResourceNotFoundError
+from azure.storage.queue import QueueClient, QueueMessage
 from dotenv import load_dotenv
 from dask.distributed import Client
 
@@ -18,10 +19,10 @@ from src.pipeline.database import (
     insert_metric_rag,
     insert_schools_and_trusts_and_local_authorities,
     insert_non_financial_data,
-    insert_financial_data
+    insert_financial_data,
 )
 
-from src.pipeline.rag import compute_rag
+from src.pipeline.rag import compute_rag, compute_user_defined_rag
 from src.pipeline.comparator_sets import (
     compute_comparator_set,
     prepare_data,
@@ -111,7 +112,7 @@ def pre_process_ks4(set_type, year) -> pd.DataFrame:
     return ks4
 
 
-def pre_process_academy_ar(set_type, year) -> (pd.DataFrame, pd.DataFrame):
+def pre_process_academy_ar(set_type, year) -> tuple[pd.DataFrame, pd.DataFrame]:
     logger.info("Processing Academy AR Data")
     academy_ar_data = get_blob(raw_container, f"{set_type}/{year}/academy_ar.xlsx")
     aar = prepare_aar_data(academy_ar_data)
@@ -179,9 +180,7 @@ def pre_process_maintained_schools_data(set_type, year, data_ref) -> pd.DataFram
     )
 
     links_data = get_blob(
-        raw_container,
-        f"{set_type}/{year}/gias_all_links.csv",
-        encoding="cp1252"
+        raw_container, f"{set_type}/{year}/gias_all_links.csv", encoding="cp1252"
     )
 
     maintained_schools = build_maintained_school_data(
@@ -340,7 +339,14 @@ def compute_comparator_set_for(data_type, mix_type, set_type, year, data):
     insert_comparator_set(set_type, mix_type, year, result)
 
 
-def compute_comparator_sets(set_type, year):
+def compute_comparator_sets(set_type: str, year: int) -> float:
+    """
+    Determine Comparator Sets.
+
+    :param set_type: the triggering message type
+    :param year: financial year in question
+    :return: duration of calculation
+    """
     start_time = time.time()
     logger.info("Computing comparator sets")
 
@@ -350,7 +356,7 @@ def compute_comparator_sets(set_type, year):
         )
     )
 
-    ms = prepare_data(
+    maintained = prepare_data(
         pd.read_parquet(
             get_blob("pre-processed", f"{set_type}/{year}/maintained_schools.parquet")
         )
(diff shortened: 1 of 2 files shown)


data-pipeline/src/pipeline/main.py
```diff
@@ -441,20 +463,121 @@ def run_compute_rag(set_type, year):
...
+    :return: updated message payload
+    """
+    msg_payload = json.loads(msg.content)
+    try:
+        if (
+            payload := msg_payload.get("payload", {}).get("kind")
+            == "ComparatorSetPayload"
+        ):
+            user_defined_comparator_set = pd.DataFrame(
+                [msg_payload["urn"]] + payload["set"]
```
PsypherPunk, line 560: Is this correct as per existing comparator-sets? i.e. the remainder are relative to the 0th element?

data-pipeline/src/pipeline/main.py
```diff
@@ -441,20 +463,121 @@ def run_compute_rag(set_type, year):
...
+            "kind": "ComparatorSetPayload",
+            "set": [
+                "145799",
+                "142875"
+            ]
+        }
+    }
+    ```
+
+    Note: user-defined comparator sets will assume pre-processing has
```
PsypherPunk, line 543: Is this safe to assume? i.e. that the main processing will have already occurred and that a given year's pre-processed data will already exist, ready for user-defined comparator-sets?

data-pipeline/src/pipeline/main.py
```diff
@@ -441,20 +463,121 @@ def run_compute_rag(set_type, year):
...
-        msg_payload["rag_duration"] = run_compute_rag(
-            msg_payload["type"], msg_payload["year"]
-        )
+    write_blob(
+        "metric-rag",
+        f"{set_type}/{run_id}/user-defined.parquet",
+        df.to_parquet(),
+    )
+
+    insert_metric_rag(
```
PsypherPunk, line 505: The `run_type`/`set_type` variables in `insert_metric_rag()` aren't entirely clear—is this right? Currently `RunId` ends up populated with `year` so this seemed to make sense…?

data-pipeline/src/pipeline/main.py
Chris-Lorch: I think this is a safe assumption to make. Realistically I don't see why we would show a particular "years" data if we did not have that data loaded.

data-pipeline/src/pipeline/main.py
colinbull: Yep... this is correct, just for clarity you are correct the naming around these is shocking *apologies* but 

* `set_type` - repersents the type of run `default`, `custom_comparator` or `custom_data`
* `run_type` - represents whether the set being computed, is Mixed (Maintained Schools and Academies) or Unmixed separate
* `run_id` - In the default case == Year, but for this it is a FE supplied ID, and the year is assumed to be the current year @jrabbott  would be best to confirm this concretely thou.

data-pipeline/src/pipeline/main.py
colinbull: Yep.. for the minute this as safe assumption.

data-pipeline/src/pipeline/main.py
colinbull: This would need to double check with @jrabbott, for the comparator sets in the `default` run the zeroth element is indeed the target URN. I don't think @jrabbott discussed this formally.

colinbull (dismissed):

data-pipeline/src/pipeline/main.py
jrabbott: There is no validation on user defined comparator set at the moment it ensure the target is included in the set - need to add this in.

Do we need the target to be 0th element, or just that it's in the set?

PsypherPunk: Not necessarily but it's an artefact of the way comparator sets are produced elsewhere: when running through all orgs. everything is compared relative to that first element.

PsypherPunk: @jrabbott, can I just check/clarify:
* the `urn` passed in the example message is the "target"…?
* as per the above, the `set` will now include that URN…?

jrabbott: 1. Correct
2. Correct

data-pipeline/src/pipeline/main.py
PsypherPunk: Updated and documented as per discussions here and on Slack.

data-pipeline/src/pipeline/main.py
```diff
@@ -452,20 +472,129 @@ def run_compute_rag(set_type, year):
...
+
+    User-defined comparator set message _content_ will be of the form:
+
+    ```json
+    {
+        "jobId": "24463424-9642-4314-bb55-45424af6e812",
+        "type": "comparator-set",
+        "runId": "c321ef6a-3b1c-4ce2-8e32-0d0167bf2fa7",
+        "year": 2022,
+        "urn": "106057",
```
PsypherPunk, line 552: @jrabbott, should this (and the `set` values) be integers when passed here?

Looking at the parquet files in the end-to-end tests, they appear to be parsed as such—not sure if that's intended though or Pandas is being "helpful"…?

jrabbott: In the front-end service & APIs URNs are strings rather than ints, so we don't have any issues with leading zeros (same with company number).

PsypherPunk: @jrabbott, any objections if I update my code to assume integers for now? If we've seen the existing behaviour work, I'd prefer to update in line with that (and fix that in isolation if it turns out to be an issue).

jrabbott: No objections from me.

Chris-Lorch (approved):

PR #991, opened 2024-06-21
feat: add custom-data processing

PsypherPunk:
### Context

adds custom-data message processing.

[AB#213664](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/213664)

- refactor messages into an explicit `enum` (stricter typing would be nice but there's no defined "default" message format…yet).
- I've refactored the comparator-set processing to just skip over the existing logic if a `target_urn` is passed, mostly to reduce the impact of any future refactoring.
- pre-processing largely uses the existing data, mapping the message's financial data over the already-extant pre-processed data.

### Guidance to review 

I've changed `year` to `run_id` in a lot of places; `year` largely serves as a run-identifier for the default job and I'd ended up with a lot of `run_id or year` logic in places. However, I've largely left the existing pre-processing as-is.

### Checklist


data-pipeline/src/pipeline/main.py
```diff
@@ -373,75 +375,202 @@ def pre_process_data(worker_client, run_type, year):
...
+        f"{run_type}/{run_id}/maintained_schools.parquet",
+        maintained.to_parquet(),
+    )
+    write_blob(
+        "pre-processed",
+        f"{run_type}/{run_id}/all_schools.parquet",
+        all_schools.to_parquet(),
+    )
+
+    insert_non_financial_data(run_type, run_id, all_schools)
```
PsypherPunk, line 452: Nope: need to reduce this to just the `target_urn` data, as discussed.

jrabbott (approved):
Looks good to me. But would be good to get another set of eyes on it

PR #1058, opened 2024-07-02
Design tweaks continued

PeteWilliams:
**School homepage**
Changed hypen in priority banner to em dash
Adjusted margins

**General**
Updated breadcrumbs to not show the active page in the breadcrumbs, as per the GDS pattern.
Note: Had a look through the [SmartBreadcrumbs docs](https://github.com/zHaytam/SmartBreadcrumbs/wiki) and [code](https://github.com/zHaytam/SmartBreadcrumbs/blob/ba32d8293481b765cfe206a23e09fb86bd8abc82/src/BreadcrumbTagHelper.cs#L151) and couldn’t see an option to configure this properly. Setting options.ActiveLiTemplate to null causes it to use a default template, but setting it to whitespace means nothing is rendered for the active page. Feel free to reject if this is too hacky! Had to update the unit tests to reflect this change.

**Commercial resources**
Removed Audio visual solutions and Office supplies, which were broken links.
Changed ‘All frameworks expired’ to ‘No resources available’

**Custom comparator sets**
Removed disabled state of ‘Choose school’ button when reached max schools - Should be enabled but show error message when clicked, otherwise users don’t know why it’s disabled.

**Trust homepage**
Added missing link to View all spending priorities 

**Trust to trust**
Added colons in table to make it clear the label and value are connected 
Removed disabled state of ‘Choose trust’ button when reached max trusts

**Schools we chose for benchmarking**
Corrected intro text content on building tab

**Custom data**
Accordion open by default


WolfyUK (changes requested):
Looks good. Before I approve I'll see if there is a better way to achieve what you are trying to with the breadcrumbs, as advised in the Description.

WolfyUK (approved):
I can't find a better way to achieve what you have done here with the breadcrumbs without drastically changing the codebase. Nice one!

PR #1068, opened 2024-07-03
Avoid unecessary processing for custom-data

PsypherPunk:
For custom-data processing:

1. when overlaying the custom data, skip processing of datasets where the target URN isn't present (previously this was effectively _adding_ this as a new row).
2. don't produce comparator-sets for datasets where the target URN isn't present (but produce an empty DataFrame with the right columns for downstream processing).
3. skip RAG calculations for all but the target URN (the comparators will be reduced as per the above but the input data is still there in its entirety; all rows need to be there for the RAG calculation but the final dataset reduced to just the relevant row).
    - skip RAG calculations entirely for datasets where the target URN isn't present (again, producing an empty, consistent DataFrame for downstream processing).

### DIFF
--- a/data-pipeline/src/pipeline/comparator_sets.py
+++ b/data-pipeline/src/pipeline/comparator_sets.py
@@ -425,7 +425,8 @@ def compute_comparator_set(
     Perform comparator-set derivation, restricted by "phase".
 
     If `data` is "custom" data, the resulting output will be restricted
-    to just the `target_urn`.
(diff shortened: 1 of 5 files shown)


jrabbott:
LGTM but would appreciate extra eyes on it.

colinbull (approved):

Chris-Lorch (approved):

PR #1094, opened 2024-07-08
fix: zero-out and recalculate custom-data

PsypherPunk:
Some further processing is needed on custom-data:

- _all_ central-services columns must be set to zero for the row in question, not just those present in the custom-data;
- per-unit and total columns must be set to zero for the row in question but _only_ for those present in the custom-data;
- the processing done in `mappings.py#map_cost_series()` must be repeated to recalculate per-unit and total costs.


jrabbott (approved):
We need to look at the linting / formatting rules post private beta, if we stay with the python pipeline.

PR #1097, opened 2024-07-08
Prototype/versioning

PeteWilliams:
Added versioning to the prototype.

Now able to archive previous versions of the prototype and have them live alongside the latest version.


jrabbott:
Still in code freeze

jrabbott (changes requested):
It feels like it could be large overhead, and with a risk of archived pages potentially breaking.

Would you be able to add some instructions into the readme, in the prototype folder, detailing how someone would go about versioning a single page?

jrabbott:
Nice one 👍 

I'll update the documentation build pipeline to include the prototype guide, so it'll produce a PDF along side all the other docs.

jrabbott (approved):

PR #1117, opened 2024-07-16
feat: adds custom rule to waf policy

jrabbott:
### Context
[AB#219080](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/219080) [AB#219081](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/219081)

Updates WAF policy;

- Adds configuration for mode in different environments
- Adds custom rule to block non uk traffic

### Guidance to review 
The expectation is that the availability test would for the 3 non-uk regions in t01

### Checklist
- [ ] ~~Your code builds clean without any errors or warnings~~
- [ ] ~~You have tested by running locally~~


web/terraform/variables.tf
```diff
@@ -58,6 +61,7 @@ variable "configuration" {
       zone_balancing_enabled = false
       worker_count           = 1
       front_door_sku_name    = "Standard_AzureFrontDoor"
+      waf_mode               = "Prevention"
```
PsypherPunk, line 64: How confident are we that no legitimate user has an IP that Azure thinks is outside the UK?

PsypherPunk: Discussed offline; marking as resolved.

`block-invalid-request-method` added to ideally block malicious traffic before reaching this rule.

PR #1119, opened 2024-07-16
Only enable optional cookies to be created on client if they have accepted the choice

WolfyUK:
### Context
[AB#219231](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/219231) [AB#218415](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/218415)

The `ai_` cookies were being created before the user had made a decision as to allow tracking cookies or not. Now they will only be created if they are accepted, but may still be removed if the user subsequently changes their mind,

### Guidance to review 
To test locally the `APPINSIGHTS_INSTRUMENTATIONKEY` environment variable needs to be set.

### Checklist


web/src/Web.App/ViewComponents/AnalyticsViewComponent.cs
```diff
@@ -14,7 +14,12 @@ public IViewComponentResult Invoke()
         }
 
         var cookiePolicy = HttpContext.Request.Cookies[Constants.CookieSettingsName];
-        var vm = new AnalyticsViewModel(instrumentationKey, cookiePolicy != "disabled");
+        var vm = new AnalyticsViewModel(instrumentationKey, cookiePolicy switch
+        {
+            "enabled" => true,
+            "disabled" => false,
+            _ => null
```
jrabbott, line 21: Should this be false, rather than null? So that it's always disabled if it falls through.

WolfyUK: On the other side I do a check for `== true` which has the same effect. Happy to switch over if you prefer

jrabbott: I can't remember off the top of my head - doesn't null or undefined with double equals return true?

WolfyUK: Yes, in JavaScript if you just perform a reference check. I perform a [value check in the Razor view](https://github.com/DFE-Digital/education-benchmarking-and-insights/pull/1119/files/e7f5167371b43b6697ffea48f2e89a84eaaf8977#diff-395a70c272e43a5b541cc6d687b29b76f9ef1857716f368db8f0446225d4c4e5R24).

WolfyUK: If it's not clear I'll just swap over

WolfyUK: Reverted to avoid confusion

web/src/Web.App/ViewComponents/AnalyticsViewComponent.cs
```diff
@@ -14,7 +14,7 @@ public IViewComponentResult Invoke()
         }
 
         var cookiePolicy = HttpContext.Request.Cookies[Constants.CookieSettingsName];
-        var vm = new AnalyticsViewModel(instrumentationKey, cookiePolicy != "disabled");
+        var vm = new AnalyticsViewModel(instrumentationKey, cookiePolicy == "enabled");
```
jrabbott, line 17: Nice one!

jrabbott (approved):

PR #1142, opened 2024-07-26
adds clickable bar chart for trust home school list

J05h-L:
### Context
[AB#217308](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/217308) - [AB#220879](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/220879)

- Adds functionality to navigate from the school lists bar charts directly to the school spending page at the first cost category of that priority type (ie high, medium, low)
- Refactors both the school spending page and the custom data version to use a partial for each category.
- Also adds CSS to add a min height to the container as without this the page will update after the user has navigated causing the desired element to be out of view.

### Guidance to review 
N/A

### Checklist

### DIFF
--- a/front-end-components/src/index.css
+++ b/front-end-components/src/index.css
@@ -211,7 +211,7 @@
 }
 
 .composed-container {
-  min-height: 200px;
+  min-height: 335px;
 }
 
 .composed-chart-wrapper {
--- a/front-end-components/src/index.css
+++ b/front-end-components/src/index.css
@@ -210,6 +210,10 @@
   color: inherit;
 }
 
+.composed-container {
+  min-height: 200px;
+}
+
 .composed-chart-wrapper {
   height: 100%;
 }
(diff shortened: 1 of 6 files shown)


web/src/Web.App/Views/SchoolSpending/_Costs.cshtml
```diff
@@ -0,0 +1,45 @@
+@using Web.App.Extensions
+@using Newtonsoft.Json
+@using Web.App.ViewModels
+@using Web.App.Domain
+
+@model CostsViewModel
+
+@if (Model.Costs.Any())
+    {
+        <a id=@Model.Id></a>
```
WolfyUK, line 10: Empty DOM element is generally not a good idea. Perhaps move the `id` attribute to the `<div>` on line `23` if this is the first `category` in the loop?

front-end-components/src/index.css
```diff
@@ -210,6 +210,10 @@
   color: inherit;
 }
 
+.composed-container {
+  min-height: 200px;
```
WolfyUK, line 214: Unfortunately it looks like this client rendered component also includes the three stat boxes as well as the chart itself, so this will need to be increased slightly, to something like `335px`, which should also cover the responsiveness.

image

J05h-L: updated in the latest commit

WolfyUK (approved):

PR #1258, opened 2024-08-30
fix: use BFR SOFA for pupil figures

PsypherPunk:
### Context

Update pupil figures as per recent comments on the ticket ("should be from…line 999 and the Y1P2 column").

[AB#217947](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/217947)

As per `Trust Revenue reserve`, derive the `Total pupils in trust` from the BFR SOFA file(s).

### Guidance to review

N/A

### DIFF
--- a/data-pipeline/src/pipeline/main.py
+++ b/data-pipeline/src/pipeline/main.py
@@ -327,6 +327,8 @@ def pre_process_bfr(run_type, year):
     )
 
     # Conditionally read in academy/SOFA files, skipping unnecessary data…
+    # TODO: "Company Reg…" isn't referenced for historic data; we can drop the
+    # join to historic academies data and just use the historic SOFA data.
     academies_y2 = None
     bfr_sofa_year_minus_two = None
     if academies_y2_file := try_get_blob(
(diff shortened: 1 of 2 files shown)


data-pipeline/src/pipeline/main.py
```diff
@@ -353,14 +353,15 @@ def pre_process_bfr(run_type, year):
...
                 )
                 .rename(
                     {"TrustUPIN": "Trust UPIN"},
                     axis=1,
                 )
-                .query("EFALineNo == 430")
+                .query("EFALineNo in (430, 999,)")
             )
 
     academies_y1 = None
```
PsypherPunk, line 367: @Chris-Lorch, IIRC when we first looked at this change we left the academies/SOFA join in place to retain the "Company Registration Number" which isn't in the latter).

However, following these historical data through: do we actually use that column from these y1/y2 DataFrames? Probably out of scope for this fix but this whole section could be simplified somewhat if that's the case.

jrabbott (dismissed):

jrabbott (approved):

Chris-Lorch (approved):

PR #1326, opened 2024-09-02
Data-pipeline: handle part-year submissions for maintained schools

PsypherPunk:
### Context

Partial-year submission for maintained schools need to be omitted from comparator-set/RAG calculations under certain circumstances.

[AB#224500](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/224500)

1. extend the maintained school data to include indicators which determine whether comparator-sets are calculated
2. in the comparator-set workflow, use these to skip comparator-set calculations.
3. moving some existing pre-processing code to `config.py` for re-use when identifying columns for the above.

### Guidance to review 

See the checks/outcomes in the linked ticket.

### DIFF
--- a/data-pipeline/src/pipeline/comparator_sets.py
+++ b/data-pipeline/src/pipeline/comparator_sets.py
@@ -226,6 +226,7 @@ def select_top_set_urns(
     boarding: np.array,
     regions: np.array,
     distances: np.ndarray,
+    include: np.array,
     base_set_size=60,
     final_set_size=30,
 ) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


Chris-Lorch (approved):

jrabbott (approved):

data-pipeline/src/pipeline/comparator_sets.py
```diff
@@ -349,13 +298,79 @@ def select_top_set_urns(
...
+    For partial-year data, financial and pupil data must be present.
+
+    For full-year data, pupil comparator sets are always generated.
+
+    :param row: grouped data
+    :return: whether to generate pupil comparator set
+    """
+    if not row["Partial Years Present"]:
+        return True
+
```
PsypherPunk, line 314: I think this (and the buildings variant below) will need a couple of slight changes:

- if those columns aren't present, `return True` (as the academy data won't have these yet)
- if those columsn _are_ present, we need to check for `not False` rather than rely on truth-y (i.e. the all-schools/mixed data will have the columns by virtue of a `join()` but they'll be null)

PsypherPunk:
@jrabbott, @Chris-Lorch: nuked your approvals but this should be good to go now.

Just need to remove that `schools` mask before merging though; PR incoming…

Chris-Lorch (approved):

PR #1349, opened 2024-09-11
DB changes for part-year Trusts

PsypherPunk:
### Context

As per the ticket, Trust financial data now differs from current Academy calculations insofar as transitioning academies need to be accounted for in both the pre- and post-transition Trusts.

[AB#226428](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/226428)

### Change proposed in this pull request

- new `TrustFinancial` table, derived from the structure of the existing `Financial` table.
  - intended to store the same data, aggregated across all component Academies for each Trust.
- updated VIEW logic, deriving data from the above table instead of existing `Financial` table:
  - `TrustBalanceHistoric`
  - `TrustBalance`
  - `TrustExpenditureHistoric`
  - `TrustExpenditure`
  - `TrustIncomeHistoric`
  - `TrustIncome`

### Guidance to review 

- are there any other Trust-related VIEWs which have been missed?

### Checklist (add/remove as appropriate)
- [x] Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)
- [x] Your code builds clean without any errors or warnings
- [x] You have run all unit/integration tests and they pass
- [x] Your branch has been rebased onto main
- [x] You have tested by running locally

PR #1354, opened 2024-09-12
feat: update pipeline to process Trust financials

PsypherPunk:
### Context

As per the ticket, Trust financial data now differs from current Academy calculations insofar as transitioning academies need to be accounted for in both the pre- and post-transition Trusts.

[AB#226428](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/226428)

Extend the pipeline to process and store Trust-level financial information:

- modify existing `prepare_aar_data()` and `build_academy_data()` functions to optionally skip the step where transitioning academies are dropped
- wrap these in new Trust-related functions to handle the Trust data
- persist data to both blob storage and the DB, the latter as per #1349.

### Guidance to review 

Ideally, it would be possible to use the raw `aar.csv` and `aar_cs.csv` files directly but:

1. some data exist elsewhere (e.g. pupil-figures are taken from `census.csv`): it's not clear what the simplest form of the input data is
2. it's not clear which calculations for the existing Academy data are required and which can be skipped

…hence the reasoning for just re-using the existing code.


data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -550,9 +546,36 @@ def prepare_aar_data(aar_path, current_year: int):
 
     aar["Is PFI"] = aar["PFI School"].map(lambda x: x == "PFI School")
 
+    # TODO: any processing required from `pre_process_trust_aar()` below?
```
PsypherPunk, line 549: `prepare_trust_aar_data()` below can be dropped entirely but:

- is it worth removing the Trust-level processing in this function at the moment?
- should entries without a Co. No. be dropped?

PsypherPunk: - Trust-level processing removed (these are added later during Academy processing.)
- `prepare_trust_aar_data()` removed.

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -724,6 +747,11 @@ def build_academy_data(
         "Total Internal Floor Area"
     ].fillna(academies["Total Internal Floor Area"].median())
 
+    # TODO: necessary for apportionment?
```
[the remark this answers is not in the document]
PsypherPunk: Being discussed internally.

jrabbott (approved):

PR #1371, opened 2024-09-19
chore: refactor of infrastructure project

jrabbott:
### Context
Tidy up project dependencies

Separated SQL and Search abstractions, to avoid unnecessary dependencies being included. Also removes IOption dependency to simplify the solution

### Guidance to review 
N/A 


platform/tests/Platform.ApiTests/packages.lock.json
```diff
@@ -565,8 +565,8 @@
       },
       "Microsoft.Bcl.AsyncInterfaces": {
         "type": "Transitive",
-        "resolved": "6.0.0",
-        "contentHash": "UcSjPsst+DfAdJGVDsu346FX0ci0ah+lw3WRtn18NUwEqRt70HaOQ7lI72vy3+1LxtqI3T5GWwV39rQSrCzAeg=="
+        "resolved": "5.0.0",
```
WolfyUK, line 568: chore: These seem to have been downgraded everywhere. I'm not sure which dependency change has caused this to occur, but it might be worth checking.

WolfyUK:
This is a chunky one! All looks good, and runs/tests fine locally. The failing checks should be resolved when other PRs are in.

platform/src/abstractions/Platform.Search/SearchConnection.cs
```diff
@@ -3,14 +3,22 @@
...
 
-namespace Platform.Infrastructure.Search;
+namespace Platform.Search;
+
+public interface ISearchConnection<T>
+{
+    Task<(long? Total, IEnumerable<ScoreResponse<T>> Response)> SearchAsync(string? search, string? filters, int? size = null);
+    Task<T> LookUpAsync(string? key);
+    Task<SearchResponse<T>> SearchAsync(SearchRequest request, Func<FilterCriteria[], string?>? filterExpBuilder = null, string[]? facets = null);
+    Task<SuggestResponse<T>> SuggestAsync(SuggestRequest request, Func<string?>? filterExpBuilder = null,string[]? selectFields = null);
```
WolfyUK, line 13: chore: Formatting issue should get picked up with a `dotnet format`

platform/tests/Platform.ApiTests/packages.lock.json
jrabbott: Good spot. I'll check that.

platform/src/abstractions/Platform.Search/SearchConnection.cs
jrabbott: Sorted

platform/tests/Platform.ApiTests/packages.lock.json
jrabbott: It's an implicit reference, which has result in the downgrade in a few places having sorted the explicit references.

PR #1372, opened 2024-09-19
feat: handle Academies part-year data

PsypherPunk:
### Context

Much as previously done with for Maintained Schools, perform various checks as per the referenced ticket and, where necessary, skip RAG calculations.

[AB#226428](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/226428)

- refactor existing Maintained-school functions for dual-use
- add functions to derive the various "check" values
- extend comparator-set checks to include Academy-specific checks

### Guidance to review

…

### DIFF
--- a/data-pipeline/src/pipeline/part_year/common.py
+++ b/data-pipeline/src/pipeline/part_year/common.py
@@ -1,0 +1,45 @@
+import pandas as pd
+
+from src.pipeline import config
+
+
+def map_has_pupil_comparator_data(df: pd.DataFrame) -> pd.DataFrame:
+    """
+    Whether the data is sufficiently populated to create a pupil
+    comparator group.
+
+    Specifically, this is the FSM and SEN data.
+
+    :param df: academy/school data
+    :return: updated DataFrame
+    """
+    pupil_comparator_columns = list(
+        set(list(config.census_column_map.values()) + config.sen_generated_columns)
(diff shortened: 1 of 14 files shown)


data-pipeline/src/pipeline/part_year/common.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Note: "Total Internal Floor Area" is specifically excluded as this
+    value is previously set to the _median_ if absent.
+
+    :param maintained_schools: maintained schools data
+    :return: updated DataFrame
+    """
+    building_comparator_columns = [
+        column for column in config.cdc_generated_columns
+        if column != "Total Internal Floor Area"
```
PsypherPunk, line 44: As documented in the docstring, this is a workaround as this gets a default value in `build_academy_data()` if it's NULL.

Thinking about it, could that actually be removed as it later gets the same default value in `prepare_data()` before comparator-sets are determined. Need to figure out if it's needed before that.

PsypherPunk: Note: this can't be removed as per the above comment as that default value is required for apportionment.

jrabbott (approved):

jrabbott (approved):

PR #1386, opened 2024-09-25
feat: deprecate academies_master_list.csv file

PsypherPunk:
### Context

Removing any reliance on the `academies_master_list.csv` file in the data-pipeline.

[AB#229900](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/229900)

- remove dependency on this file from `build_academy_data()`
- add fields to input fields for AAR data to cover those missing
- update phase-mapping to accommodate potentially fewer sources

### Guidance to review 

I'll add specific comments to any parts of note.


data-pipeline/src/pipeline/input_schemas.py
```diff
@@ -321,11 +300,13 @@
     "BNCH21606 (Agency supply teaching staff)": "float",
     "BNCH21403 (Energy)": "float",
     "BNCH21402 (Water and sewerage)": "float",
+    "Valid To": "string",
```
PsypherPunk, line 303: Although this has a `Valid From` counterpart, as per the `academy_master_list` above, it doesn't appear to be used anywhere.

data-pipeline/src/pipeline/mappings.py
```diff
@@ -236,8 +239,8 @@ def map_cost_series(category_name, df, basis):
 
     for sub_category in sub_categories:
         df[sub_category + "_Per Unit"] = df[sub_category].fillna(0) / basis
-        df[sub_category + "_Per Unit"].replace(
-            [np.inf, -np.inf, np.nan], 0, inplace=True
+        df[sub_category + "_Per Unit"] = df[sub_category + "_Per Unit"].replace(
```
PsypherPunk, line 242: This is just to fix a warning coming from Pandas (of which there are a few more to fix elsewhere):

```sh
…/education-benchmarking-and-insights/data-pipeline/src/pipeline/mappings.py:242: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.
```

data-pipeline/src/pipeline/mappings.py
```diff
@@ -14,11 +14,14 @@ def map_ofsted_rating(rating: str):
...
 
-def map_phase_type(establishment_code: int, phase_code: int, provision: str):
-
-    if (pd.isna(establishment_code) or pd.isna(phase_code) or pd.isna(provision)):
+def map_phase_type(
+    establishment_code: int,
+    phase_code: int,
+    provision: str | None = None,
+):
+    if pd.isna(establishment_code) or pd.isna(phase_code) or pd.isna(provision):
```
PsypherPunk, line 22: Left in place as it's out-of-scope but is this correct?

If _any_ of them are null, return null? If _all_ are null, I could see that but wouldn't the preference be to rely on at least one of them?

data-pipeline/src/pipeline/mappings.py
```diff
@@ -14,11 +14,14 @@ def map_ofsted_rating(rating: str):
             return rating
 
 
-def map_phase_type(establishment_code: int, phase_code: int, provision: str):
-
-    if (pd.isna(establishment_code) or pd.isna(phase_code) or pd.isna(provision)):
+def map_phase_type(
+    establishment_code: int,
+    phase_code: int,
+    provision: str | None = None,
```
PsypherPunk, line 20: Discussed elsewhere but we'll need to validate this: the academies will only have the GIAS-sourced first two values so this will default to `None` in those instances.

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -807,10 +789,17 @@ def build_academy_data(
     academies["Email"] = ""
     academies["HeadEmail"] = ""
 
+    academies = academies.merge(
```
PsypherPunk, line 792: `Trust UPIN` is now pulled in from the Central Services file so the ordering of the academies/central-services/trusts* merging has changed here.

*this should probably be removed entirely when the part-year-trusts work is done.

data-pipeline/tests/unit/pre_processing/conftest.py
```diff
@@ -309,6 +311,7 @@ def aar_central_services_data() -> pd.DataFrame:
             "BNCH21606 (Agency supply teaching staff)": [1000, 1001, 1002],
             "BNCH21403 (Energy)": [1000, 1001, 1002],
             "BNCH21402 (Water and sewerage)": [1000, 1001, 1002],
+            "BNCHBAI061 (Coronavirus Govt Funding)": [1000, 1001, 1002],
```
PsypherPunk, line 314: I'm stumped by this one: it wasn't in this `dict` but the tests pass on `main`…?!

data-pipeline/src/pipeline/input_schemas.py
```diff
@@ -321,11 +300,13 @@
...
 }
 
 aar_central_services_index_col = "Lead_UPIN"
 aar_central_services = {
     "Lead_UPIN": "Int64",
+    "Company_Number": "string",
```
WolfyUK, line 309: Just checking that this column is likely to be in the same format as in the `academy_master_list` and/or the potential leading `0` is dealt with somewhere else

data-pipeline/tests/unit/pre_processing/conftest.py
WolfyUK: So it's not that they didn't actually run?!

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -714,21 +703,14 @@ def build_academy_data(
...
-    )
-
-    academies_list = academies_list[~mask]
-    academies_base = academies_list.merge(
-        schools.reset_index(), left_index=True, right_on="LA Establishment Number"
-    )
-
     academies = (
-        academies_base.merge(census, on="URN", how="left")
+        aar[~aar.index.isna()]
```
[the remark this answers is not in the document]
PsypherPunk: Done: moved the stripping of extraneous rows to `prepare_aar_data()` and added a test.

data-pipeline/src/pipeline/input_schemas.py
PsypherPunk: Good shout: the `aar.csv` and `aar_cs.csv` files use their respective `Company_Number` to join and they're both in the same format for that purpose.

The `map_company_number()` call which zero-pads them _after_ that is retained (down on line 997).

data-pipeline/tests/unit/pre_processing/conftest.py
PsypherPunk: I've checked out `main` and run this test and it passes: this column _isn't_ there.

…but it _is_ there now. Need to account for where it was being dropped.

PsypherPunk: This is indeed down to a change I made:

```python
central_services_financial = pd.read_csv(
    cs_path,
    encoding="utf-8",
    usecols=lambda x: x in input_schemas.aar_central_services.keys(),
    dtype=input_schemas.aar_central_services,
)
```

…versus:

```python
central_services_financial = pd.read_csv(
    cs_path,
    encoding="utf-8",
    usecols=input_schemas.aar_central_services.keys(),
    dtype=input_schemas.aar_central_services,
)
```

The former will check whether any columns in the input file are in the expected schema and, if so, include them.

I'd argue the latter is correct, otherwise we're processing inconsistently-structured data but for the sake of confusion, I'll revert.
The latter will include columns from the expected schema and, if they're _not_ they're, still include the column.

PsypherPunk: Reverted.

jrabbott:
image

data-pipeline/src/pipeline/input_schemas.py
jrabbott: Probably something we can pick up with Ben to ensure we've defined the required input fields mappings

data-pipeline/src/pipeline/mappings.py
jrabbott: Do we know the impact across the 3 years?  (i.e. how schools in the AAR don't appear in the associated GIAS file gias, gias_all_link or gias_links)

data-pipeline/src/pipeline/mappings.py
jrabbott: Agree - I'm not sure why we need to rely on all 3.

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -776,9 +758,11 @@ def build_academy_data(
         axis=1,
     )
 
+    # TODO: remove; duplicate of `Overall Phase`, above?
```
jrabbott, line 761: Agree - assuming `SchoolPhaseType` is unused. It's not written to the db.

jrabbott (commented):
Have all the mapping updates / field name changes been covered? (I'm still seeing Company Registration Number & Academy Trust UPIN for example, but unsure on the impact)

- Company Registration Number => Company_Number
- Incorporation Date => Incorporation_Date
- Academy Trust Name =>Company_Name
- Academy Name => Academy_Name
- Academy UPIN =>ACADEMYUPIN
- Trust Type => MAT/SAT
- LA Establishment Number => LAEstab
- Date Opened => Date_Opened
- Valid From => Valid From
- Valid to => Valid To
- Date joined or opened if in period => Date joined or opened if in period:
- Date left or closed if in period => Date left or closed if in period:
- Academy Status =>ACADEMYSTATUS
- Academy Trust Status => ACADEMYTRUSTSTATUS
- Number of Academies in Trust => NUMBEROFACADEMIESINTRUST
- Number of Pupils => NUMBEROFPUPILS
- Academy Trust UPIN => Lead_UPIN

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -792,7 +776,6 @@ def build_academy_data(
             "LA (code)": "LA Code",
             "LA (name)": "LA Name",
             "Academy Trust Name": "Trust Name",
```
jrabbott, line 778: Academy Trust Name and Academy Trust UPIN will need updating based on AAR and AAR_CS

Company_Name & Lead_UPIN

data-pipeline/src/pipeline/mappings.py
PsypherPunk: As far as I can tell, there's a single academy affected. Based on the academies-master-list I have, it's not actually present; based on the data in d01 and Josh, it's there and gets populated (which I think speaks to the wider reason that file needs to go).

I flagged "phases" as something specifically to look at during QA.

data-pipeline/src/pipeline/pre_processing.py
PsypherPunk: It's used for maintained-schools (but so is `Overall Phase`) and used as part of the comparator-set processing.

In both cases, it'd probably be easy to remove but for the sake of keeping the changes to a minimum, I vote for picking up later.

data-pipeline/src/pipeline/pre_processing.py
PsypherPunk: Going to run through the data at this point and update accordingly: `URN_x` isn't there, for instance; `Trust UPIN` will appear following the merge onto `central_services` in a dozen lines or so.

data-pipeline/src/pipeline/mappings.py
jrabbott: I think this method needs rewriting. I've done a bit of analysis on the GIAS data, and the method can be simplified to the below.

We'll require the `PhaseOfEducation` for all cases, but only `TypeOfEstablishment` in two cases. (Apologies for the C#)

````
string overallPhase = phaseOfEducation switch
{
    0 => GetNotApplicablePhases(typeOfEstablishment),
    1 => "Nursery",
    2 or 3 => "Primary",
    4 or 5 => GetSecondaryPhases(typeOfEstablishment),
    6 => "Post-16",
    7 => "All-through",
    _ => "Unkown"
};

string GetSecondaryPhases(int type)
{
    string result = type switch
    {
        40 => "University Technical College",
        _ => "Secondary"
    };

    return result;
}

string GetNotApplicablePhases(int type)
{
    string result = type switch
    {
        14 => "Pupil Referral Unit",
        39 or 45 or 46 => "Special",
        38 or 42 or 43 => "Alternative Provision",
        _ => "Unknown"
    };

    return result;
}
```

data-pipeline/src/pipeline/pre_processing.py
jrabbott: Agree

data-pipeline/src/pipeline/mappings.py
jrabbott: @J05h-L is there any reason why the above wouldn't work for the Overall Phases?

data-pipeline/src/pipeline/pre_processing.py
PsypherPunk: - `URN_x` and `UKPRN_x` removed (no longer present)
- `Company_Name` -> `Trust Name` moved to `prepare_aar_data`
- `config.income_category_map["academies"]` and ` config.cost_category_map["academies"]` removed as this already takes place earlier in `prepare_aar_data`

data-pipeline/src/pipeline/mappings.py
J05h-L: I don't think this would account for PhaseOfEducation (code) == 0 and TypeOfEstablishment (code) not in our checks. Then we should use Type of Provision - Phase from the academy_master_list or OverallPhase from the maintained_schools_master_list which we won't here

PsypherPunk: @jrabbott, updated to reflect the above.

data-pipeline/tests/unit/pre_processing/test_maintained_schools.py
```diff
@@ -103,12 +101,13 @@ def test_map_school_attrs():
     assert {"Email", "HeadEmail", "Trust Name"}.issubset(list(actual.columns))
 
 
+@pytest.mark.skip("TODO: needs updating to reflect phase-type mappings.")
```
PsypherPunk, line 104: I've skipped any phase-type-related tests for now: these should, however, be updated ASAP when we're confident in the changes.

data-pipeline/src/pipeline/mappings.py
jrabbott: LGTM

data-pipeline/src/pipeline/mappings.py
```diff
@@ -14,46 +14,45 @@ def map_ofsted_rating(rating: str):
...
+    establishment_code: int,
+    phase_code: int,
+) -> str:
     match phase_code:
-        case 7:
-            return "All-through"
+        case 0:
+            return _map_not_applicable_phases(establishment_code)
+        case 1:
+            "Nursery"
```
jrabbott, line 45: Missing return

jrabbott (approved):

data-pipeline/tests/unit/pre_processing/test_maintained_schools.py
PsypherPunk: Restored.

jrabbott (approved):

PR #1389, opened 2024-09-26
Assign CalVer to each pipeline run and prevent unexpected version bumps

WolfyUK:
### Context
[AB#230056](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/230056) [AB#229406](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/229406)

Mostly handled via updated `vars.yml` template, but the addition of `name`, even if only temporary until elsewhere in each pipeline, prevents the version bump. This includes the `production` pipeline, which will also make it through to App Insights annotations (albeit with the same version number for each application deployment for the time being). For `front-end-components`, the build number format is slightly different, and is distinguishable from other pipelines that also manage this version and subsequently push to the private artifact feed.

### Guidance to review 
Any pipeline may be manually triggered with this source branch to validate the non-trunk based versioning. As soon as a non-PR originated build is triggered on `main` the 'production-ready' version numbers will be used instead. This should be validated in each pipeline after this PR has been merged.


WolfyUK:
> We only want CalVer for the release pipeline, not all pipelines

Not a problem. What about `support-analytics`?

pipelines/common/vars.yaml
```diff
@@ -1,6 +1,32 @@
...
-  Version.BuildNumber: $(Version).$(Version.Revision)
-  Version.BuildVersion: $(Version.BuildNumber)
-  ShouldDeploy: $[and(eq(variables['Build.SourceBranch'], 'refs/heads/main'), ne(variables['Build.Reason'], 'PullRequest'))]
\ No newline at end of file
+  - name: ShouldDeploy
+    value: $[and(eq(variables['Build.SourceBranch'], 'refs/heads/main'), ne(variables['Build.Reason'], 'PullRequest'))]
+  - name: Version
+    value: 1.0
+  - name: Version.Revision
+    ${{ if eq(variables['ShouldDeploy'], 'true') }}:
```
jrabbott, line 20: This will always return false. The `if` is done at compile time whereas the variable value is set at runtime.

WolfyUK: Good spot - I thought I had validated this in a test pipeline but seemingly not until just now!

jrabbott (approved):

PR #1394, opened 2024-09-27
Modified `LocalAuthority` API response model to include `Schools` collection

WolfyUK:
### Context
AB#230396 AB#229325

Perform single API call to obtain Local Authority and related part-School collection rather than require multiple API calls to achieve the same result.

### Guidance to review 
Build and run Platform and Web locally and then browse to Local Authority home/census/comparison. Visually, results should be identical before and after this change.


J05h-L (approved):

jrabbott:
No action (just thinking out loud) - by using the SchoolService to query for the list of schools, I assume this result in a second db connection being opened. 

We it be simpler to not resuse the the SchoolSevice, thus not perform the second query and construct the response in the function body. However instead perform the query to get the LocalAuthoritySchools in the GetAsync method on LocalAuthorityService and build the LocalAuthority model in that service.

WolfyUK:
> I assume this result in a second db connection being opened

Yes, this is precisely the case. This work arose due to overheads observed in API, not DB requests. It is likely that a future pass will resolve the above here and elsewhere.

WolfyUK:
See AB#230534 for follow-up DB work

PR #1433, opened 2024-10-08
feat: cache response from insightApi.GetCurrentReturnYears()

J05h-L:
### Context
[AB#229461](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/229461) - [AB#230542](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/230542)

Adds in memory cache for the web app. 
Cache for response from insightApi.GetCurrentReturnYears()
Options set via config so can be amended without code change if needed.

### Guidance to review 
N/A

### DIFF
--- a/web/src/Web.App/Services/FinanceService.cs
+++ b/web/src/Web.App/Services/FinanceService.cs
@@ -1,3 +1,6 @@
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Options;
+using Web.App.Cache;
 using Web.App.Domain;
 using Web.App.Infrastructure.Apis;
 using Web.App.Infrastructure.Apis.Insight;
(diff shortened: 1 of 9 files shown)


web/src/Web.App/Services/FinanceService.cs
```diff
@@ -9,7 +12,34 @@ public interface IFinanceService
...
\ No newline at end of file
+    private readonly CacheOptions _cacheOptions = cacheOptions.Value;
+    private readonly string _cacheKey = cacheOptions.Value.CacheKey ?? throw new NullReferenceException(nameof(CacheOptions));
+
+    public async Task<FinanceYears> GetYears()
+    {
+        FinanceYears data;
+        if (memoryCache.TryGetValue(_cacheKey, out var cachedData))
+        {
+            data = cachedData as FinanceYears ?? throw new InvalidOperationException(nameof(FinanceYears));
```
WolfyUK, line 28: If the cached data is not of the expected format it would be better to just get and cache a new value rather than throw an exception. A warning log message may be helpful though.

web/src/Web.App/Services/FinanceService.cs
```diff
@@ -9,7 +12,34 @@ public interface IFinanceService
     Task<FinanceYears> GetYears();
 }
 
-public class FinanceService(IInsightApi insightApi) : IFinanceService
+public class FinanceService(
+    IInsightApi insightApi,
+    IMemoryCache memoryCache,
+    IOptions<CacheOptions> cacheOptions) : IFinanceService
```
WolfyUK, line 18: Would it be possible to type the options as `ReturnYearsCacheOptions` or similar, inheriting from `CacheOptions`? Then in the service configuration you can be explicit about which config section should be used to resolve the values in the case of multiple cache config sections in the future.

web/src/Web.App/Services/FinanceService.cs
```diff
@@ -9,7 +12,34 @@ public interface IFinanceService
...
 {
-    public async Task<FinanceYears> GetYears() => await insightApi.GetCurrentReturnYears().GetResultOrThrow<FinanceYears>();
-}
\ No newline at end of file
+    private readonly CacheOptions _cacheOptions = cacheOptions.Value;
+    private readonly string _cacheKey = cacheOptions.Value.CacheKey ?? throw new NullReferenceException(nameof(CacheOptions));
```
WolfyUK, line 21: Maybe `ArgumentNullException` instead?

web/src/Web.App/Services/FinanceService.cs
J05h-L: thanks, updated to do this in latest commits

web/src/Web.App/Services/FinanceService.cs
J05h-L: Amended CacheOptions so it can be used as suggested

web/src/Web.App/Services/FinanceService.cs
J05h-L: removed the exception, defaults set where values are null

jrabbott (approved):

PR #1446, opened 2024-10-10
bugfix: year mismatch to return data

jrabbott:
Removes dependency on year var to calculate various fields.

Pipeline run year won't always match the return year data.

### Guidance to review 
Pipeline ran successfully locally with correct output


data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -740,27 +736,11 @@ def build_academy_data(
         axis=1,
     )
 
-    academies["Status"] = academies.apply(
```
PsypherPunk, line 743: Is `Status` just never used throughout?

data-pipeline/src/pipeline/mappings.py
```diff
@@ -205,40 +140,70 @@ def map_is_surplus_deficit(closing_balance: float):
...
-    :return: number of months between start and end dates
-    """
-    if opening:
-        if d2.month == 9:
-            return 12 if d2.day <= 10 else 11
-        else:
-            return (d1.year - d2.year) * 12 + d1.month - d2.month + 1
-
-    return (d1.year - d2.year) * 12 + d1.month - d2.month
+def _period_from_start(start):
```
PsypherPunk, line 143: Nitpick: could either benefit from a docstring or perhaps just calling it `_months_from_start()` would make it clearer…?

PsypherPunk (commented):
Couple of clarifying points/questions but largely LGTM.

data-pipeline/src/pipeline/mappings.py
PsypherPunk: Actually, this one confused me: I though this was "period-from-the-start-of-the-academic-year" (but it's "period-from-this-start-date-to-the-end-of-the-academic-year").

Maybe `_months_left_in_academic_year`/`_months_spent_in_academic_year` or similar might be clearer…?

data-pipeline/src/pipeline/pre_processing.py
jrabbott: From a code search I couldn't find any usages. Pipeline also ran successfully.

data-pipeline/src/pipeline/mappings.py
jrabbott: Only think to consider, this is a specific calc for the academy return. The dates align to the AAR return period (which is the academic year), it would be a different calc for maintained schools.

jrabbott: @PsypherPunk methods renamed

data-pipeline/src/pipeline/pre_processing.py
```diff
@@ -806,26 +802,11 @@ def build_academy_data(
...
-            pd.to_datetime(df["OpenDate"]),
-            pd.to_datetime(df["CloseDate"]),
-            pd.to_datetime(accounts_return_period_start_date),
-            pd.to_datetime(academy_year_start_date),
-            pd.to_datetime(academy_year_end_date),
-        ),
-        axis=1,
-    )
-
+    # TODO: should factor into apportionment.
```
PsypherPunk, line 805: Think this might be an artefact of a merge/rebase? This TODO was removed recently, IIRC.

PsypherPunk (approved):

PR #1506, opened 2024-10-29
Doc/create azure devops test plan for release

Faizan-ahmad00:
Documentation on how to create and execute test plan for release which will cover the smoke testing as well.


jrabbott (changes requested):
Having read this, I think we are miss using Test plans, suites and cases in Azure Devops.

Faizan-ahmad00:
> Having read this, I think we are miss using Test plans, suites and cases in Azure Devops.

updated the plan as per discussion.

jrabbott (approved):
Looks good! Nice one.

PR #1509, opened 2024-10-29
Tech debt/add test data in repo

Faizan-ahmad00:
updated test data in the folder
updated the seeding document


documentation/quality-assurance/6_Data-Seeding-in-Test-Environment.md
jrabbott, line 1: I think this document might be better placed in the test-data folder with the data files? Not sure 100%, what do you think?

test-data/Parameters.csv
jrabbott, line 1: Not sure if we should be import this file or not? What are your thoughts? It's currently missing the BFR parameter.

Faizan-ahmad00: Since we currently don’t have BFR test data in d02, the BFR parameter is missing from the file. However, as the data is up to 2022, I think it's still worthwhile to import the file as it stands. We can update the file when we update to new data(2023/2024) or add BFR tests as they become available in future updates.

documentation/quality-assurance/6_Data-Seeding-in-Test-Environment.md
Faizan-ahmad00: ammm I see your point about placing the document with the data files. it would be helpful to have this information easily accessible when updating data. 
Updating the readme with the content and removing this file.

jrabbott (approved):

PR #1554, opened 2024-11-06
fix: limit Container App name to 32 chars.

PsypherPunk:
### Context

`apply` is failing as a _"name must…not be more than 32 characters."_

[AB#235428](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/235428)

- drop the `data` part of `data-pipeline`
- truncate the name to 32 characters in any case to avoid future issues.

### Guidance to review 

Again, no failures in the `plan` just the `apply`.

- [ ] ~You have run all unit/integration tests and they pass~
- [ ] ~You have tested by running locally~

### DIFF
--- a/data-pipeline/terraform/container_app/main.tf
+++ b/data-pipeline/terraform/container_app/main.tf
@@ -1,5 +1,5 @@
 resource "azurerm_container_app" "data-pipeline" {
-  name                         = "${var.environment-prefix}-ebis-data-pipeline-${var.container-app-name-suffix}"
+  name                         = local.container-app-name
   container_app_environment_id = var.container-app-environment-id
   resource_group_name          = var.resource-group-name
   revision_mode                = "Single"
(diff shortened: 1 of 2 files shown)


data-pipeline/terraform/container_app/main.tf
```diff
@@ -1,5 +1,5 @@
 resource "azurerm_container_app" "data-pipeline" {
-  name                         = "${var.environment-prefix}-ebis-data-pipeline-${var.container-app-name-suffix}"
```
WolfyUK, line 2: How close was the original to the `32` limit? Is the `suffix` going to be unduly truncated and/or should a different naming convention be defined?

PsypherPunk: `s198d01-ebis-data-pipeline-default` was being truncated to `s198d01-ebis-data-pipeline-defau` so it's only a couple of characters; `custom` would be less so.

I've dropped the `data` part which should avoid truncation in this instance. I've added the `substr()` call "just in case".

WolfyUK (approved):

PR #1603, opened 2024-11-20
Create and insert Mean as well as Median (with diffs) into `MetricRAG` table

WolfyUK:
### Context
[AB#238511](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/238511) [AB#238512](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/238512) [AB#236925](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/236925)

- Added new columns to `MetricRAG` table
- During pipeline run, insert mean and median to correct `MetricRAG` columns

### Guidance to review 
A future piece of work will deprecate the `Mean` columns. For now, both averages will exist so as to not break the service. This can be validated by running the `data-pipeline` locally, after having first run the database migrations.

### DIFF
--- a/data-pipeline/src/pipeline/rag.py
+++ b/data-pipeline/src/pipeline/rag.py
@@ -121,11 +121,13 @@ def category_stats(urn, category_name, data, ofsted_rating, rag_mapping, close_c
 
     percentile = find_percentile(series, value)
     decile = int(percentile / 10)
-    mean = np.median(series)
-    diff = value - mean
+    mean = np.mean(series)
+    diff_mean = value - mean
+    median = np.median(series)
+    diff_median = value - median
     diff_percent = (
-        (diff / mean) * 100
-        if mean != 0 and mean != np.inf and mean != np.nan and not pd.isna(mean)
+        (diff_median / median) * 100
+        if median != 0 and median != np.inf and median != np.nan and not pd.isna(median)
         else 0
     )
     cats = category_name.split("_")
@@ -135,8 +137,11 @@ def category_stats(urn, category_name, data, ofsted_rating, rag_mapping, close_c
         "SubCategory": cats[1],
         "Value": value,
         "Mean": mean,
-        "DiffMean": diff,
+        "DiffMean": diff_mean,
+        "Median": median,
+        "DiffMedian": diff_median,
         "Key": key,
+        # note: `PercentDiff` is not consumed in api or web
         "PercentDiff": diff_percent,
         "Percentile": percentile,
         "Decile": decile,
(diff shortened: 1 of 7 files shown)


J05h-L (approved):

data-pipeline/src/pipeline/rag.py
```diff
@@ -121,23 +121,32 @@ def category_stats(urn, category_name, data, ofsted_rating, rag_mapping, close_c
...
+    diff_mean = value - mean
     diff_percent = (
-        (diff / mean) * 100
+        (diff_mean / mean) * 100
         if mean != 0 and mean != np.inf and mean != np.nan and not pd.isna(mean)
         else 0
     )
+    median = np.median(series)
+    diff_median = value - median
+    diff_percent = (
```
jrabbott, line 133: `diff_percent` has me a little confused. Seems like it is being assigned based on the mean and then again based on the median.

Feels like we either need both diff_percent_mean & diff_percent_median, or just one which is based on either mean or median?

jrabbott (commented):
This approach needs careful consideration, because current the web & API current display the median (but labelled as mean) as required.

By implementing this way there is a risk that the Web/API will return and display the mean, when it should be displaying the median.

data-pipeline/src/pipeline/rag.py
WolfyUK: I added a comment a little further down to say that value wasn't being consumed anywhere on the front end, so this distinction was not of concern. Happy to add another field if you feel it is necessary to persist this value.

WolfyUK:
> By implementing this way there is a risk that the Web/API will return and display the mean, when it should be displaying the median.

Agreed. It was hoped the changes elsewhere in the repo would follow shortly afterwards to reduce the impact, especially before a new pipeline run is completed. If it is preferred, the API and Web changes can all be done as part of a single PR.

jrabbott:
My preference would be for the changes to be in separate PRs. 

It was more the QA impact and awareness of the scope of these changes. There might have been an assumption that the prop/field would just be renamed therefore no risk of showing the wrong value (the code would break or show nothing).

QA will need to be aware of all places where these values are surfaced.

jrabbott (approved):

WolfyUK:
> There might have been an assumption that the prop/field would just be renamed

Perhaps that is the wider question: should the 'average' be 'median' _everywhere_ for RAGs and 'mean' dropped completely, or is there a case where one or the other may be consumed? I was trying to be cautious with the implementation, but if there is no need for both then the follow up work to clean this up should follow swiftly afterwards.

jrabbott:
`Median` should be used everywhere at the moment. However I'm not sure if `mean` would have any value in future. 



https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-rename-transact-sql?view=sql-server-ver16

WolfyUK:
Closing in anticipation of alternative implementation to simply rename the original column rather than support both `median` and `mean`.

PR #1749, opened 2025-01-08
build: add pytest-randomly

PsypherPunk:
Make use of [`pytest-randomly`](https://pypi.org/project/pytest-randomly/) to randomise the test order, making sure there are no implicit dependencies between tests.


jrabbott:
Failing build

PsypherPunk:
@jrabbott, it's the branch-name; will re-raise with a valid one.

PR #1753, opened 2025-01-09
feat: suppress negative and zero values for school benchmark spending charts

J05h-L:
### Context
[AB#235748](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/235748) - [AB#243321](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/243321)

In both the horizontal bar chart and table chart components a filter has been added to remove any establishment that is not the selected school and has a value <= 0 if we wish to suppress negative and zero values.

A prop is passed via the dataset attributes set in the web app view to the CompareYourCosts front end view to set context to be passed down to these components setting whether to suppress negative or zero values and the required message to be displayed as part of the warning text using the error banner component.

This is implemented for the default school benchmark spending and custom data benchmark spending pages. This is not used for the trust view school spending page so the behaviour of that page should remain untouched.

The existing but unused warning banner component has been removed

### Guidance to review 
Awaiting content design input on the content of the message


front-end-components/src/components/warning-banner/component.tsx
```diff
@@ -4,8 +4,14 @@ export const WarningBanner: React.FC<WarningBannerProps> = (props) => {
   const { isRendered, message } = props;
 
   return isRendered ? (
-    <div className="govuk-inset-text govuk-body govuk-!-margin-0">
-      {message}
+    <div className="govuk-warning-text">
```
WolfyUK, line 7: discussion: I understand that this component is no longer in use elsewhere, but I'm not sure if this design change fits well on the page. Please can this be included in the Design discussion as mentioned in the PR description?

WolfyUK: Alternatively the `<ErrorBanner />` could be used instead, which appears to now be identical to the modified `<WarningBanner />`

front-end-components/src/components/charts/table-chart/component.tsx
```diff
@@ -40,116 +45,139 @@ export const TableChart: React.FC<
...
   );
 
+  const dataPointKey = trust ? "totalValue" : "value";
+
+  const keyField = trust ? "companyNumber" : "urn";
+
+  const filteredData = useMemo(() => {
+    return data?.filter((d) =>
+      suppressNegativeOrZero
+        ? (d as never)[dataPointKey] > 0 ||
```
WolfyUK, line 55: nitpick: `never`s can be resolved by casting the keyed fields, e.g.:

```ts
  const dataPointKey = (trust ? "totalValue" : "value") as keyof (
    | SchoolChartData
    | TrustChartData
  );

  const keyField = (trust ? "companyNumber" : "urn") as keyof (
    | SchoolChartData
    | TrustChartData
  );

  const filteredData = useMemo(() => {
    return data?.filter((d) =>
      suppressNegativeOrZero
        ? (d[dataPointKey] as number) > 0 || d[keyField] === selectedSchool
        : true
    );
  }, [data, suppressNegativeOrZero, dataPointKey, keyField, selectedSchool]);
```

front-end-components/src/components/charts/horizontal-bar-chart/component.tsx
```diff
@@ -69,8 +76,25 @@ function HorizontalBarChartInner<TData extends ChartDataSeries>(
...
+    SuppressNegativeOrZeroContext
+  );
+  const selectedSchool = useContext(SelectedEstablishmentContext);
+
+  const dataPointKey = trust ? "totalValue" : "value";
+
+  const filteredData = useMemo(() => {
+    return data.filter((d) =>
+      suppressNegativeOrZero
+        ? (d as never)[dataPointKey] > 0 ||
```
WolfyUK, line 92: nitpick: `never`s can be resolved by casting the keyed fields, e.g.:

```ts
  const dataPointKey = (trust ? "totalValue" : "value") as keyof (
    | SchoolChartData
    | TrustChartData
  );

  const keyField = (trust ? "companyNumber" : "urn") as keyof (
    | SchoolChartData
    | TrustChartData
  );

  const filteredData = useMemo(() => {
    return data?.filter((d) =>
      suppressNegativeOrZero
        ? (d[dataPointKey] as number) > 0 || d[keyField] === selectedSchool
        : true
    );
  }, [data, suppressNegativeOrZero, dataPointKey, keyField, selectedSchool]);
```

WolfyUK: The above could also be tidied up in `<HorizontalBarChartWrapper />` where I was likely the original author! 😬

front-end-components/src/components/charts/table-chart/component.tsx
```diff
@@ -40,116 +45,139 @@ export const TableChart: React.FC<
     </a>
   );
 
+  const dataPointKey = trust ? "totalValue" : "value";
+
+  const keyField = trust ? "companyNumber" : "urn";
+
+  const filteredData = useMemo(() => {
```
WolfyUK, line 52: It is a little concerning that the filtering is done in two places. Is there any way this could be done in the parent component instead so that the `data` prop can be used as-is and another prop used to display the warning?

front-end-components/src/components/warning-banner/component.tsx
J05h-L: Updated to use the error banner component and removed warning banner

J05h-L: Will try arrange a design crit to cover the use and placement of the warning as well as the content

front-end-components/src/components/charts/table-chart/component.tsx
J05h-L: done!

front-end-components/src/components/charts/horizontal-bar-chart/component.tsx
J05h-L: and done!

front-end-components/src/components/charts/table-chart/component.tsx
J05h-L: My original intention was for this all to take place in `HorizontalBarChartWrapper` but after discussions I took the steer that this should be done in the components themselves as they should have the responsibility. Should allow us to use the components independently with the same behaviour.

Happy to take guidance on this.

WolfyUK (approved):

PR #1758, opened 2025-01-10
feat: suppress negative and zero values for spending priorities page

J05h-L:
### Context
[AB#235748](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/235748) - [AB#243322](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/243322) & [AB#244430](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/244430)

filter non selected schools in the comparator set when they have negative or zero values for each cost category for the spending priorites page (default and custom data) so they are not included in the charts.
display a warning for each chart when values are suppressed.

when calculating the values for the rag rating for each category in the data pipeline exclude other schools in the comparator set when they have negative or zero values so the calculations match the schools displayed on the graphs on the spending priorities page

add two new app css classes for the commercial resources container to both fix an existing issue with margin that is only required when viewing on smaller screen after the grid row collapses and to add padding to align the text when the warning is adjacent.

### Guidance to review 
Check implementation for the logic in web and data pipeline and how this is displayed at various browser sizes.

### DIFF
--- a/data-pipeline/src/pipeline/rag.py
+++ b/data-pipeline/src/pipeline/rag.py
@@ -116,7 +116,7 @@ def category_stats(urn, category_name, data, ofsted_rating, rag_mapping, close_c
     key = "outstanding" if ofsted_rating.lower() == "outstanding" else "other"
     key += "_10" if close_count > 10 else ""
 
-    series = data[category_name]
+    series = _get_series_for_category(data, category_name, urn)
     value = series.loc[urn]
 
     percentile = find_percentile(series, value)
@@ -316,3 +316,19 @@ def compute_user_defined_rag(
                     exc_info=error,
                 )
                 return
+            
+def _get_series_for_category(data: pd.DataFrame, category: str, urn: int) -> pd.Series:
+    """
+    Filters the comparator set data for the specified category, retaining only 
+    positive values or the value for the specified org. (identified by URN).
+
+    This ensures that only positive expenditure values are considered for 
+    rag calculations.
+
+    :param data: A DataFrame containing comparator data for RAG calculations.
+    :param category: The category column to filter on
+    :param urn: The URN of the org. whose value should always be included.
+    :return: A Series containing the filtered values
+    """
+
+    return data.loc[(data[category] > 0) | (data.index == urn), category]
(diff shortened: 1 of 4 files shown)


web/src/Web.App/AssetSrc/scss/main.scss
```diff
@@ -389,4 +389,20 @@ hr.govuk-section-break--print {
...
+  @include govuk-media-query('tablet') {
+    margin-top: 0;
+  }
+}
+
+.app-resources-warning-spacing {
+  padding-top: 0;
+
+  @include govuk-media-query('tablet') {
+    padding-top: 0.625rem;
```
WolfyUK, line 406: nitpick: To prevent the 'magic number' here and above you should be able to import from [GDS](https://design-system.service.gov.uk/styles/spacing/) using the appropriate mixin. e.g. `govuk-spacing()`.

WolfyUK (changes requested):

jrabbott (changes requested):
This requires a change to the data pipeline as the need to have zero and negative suppressed.

data-pipeline/src/pipeline/rag.py
```diff
@@ -116,7 +116,7 @@ def category_stats(urn, category_name, data, ofsted_rating, rag_mapping, close_c
     key = "outstanding" if ofsted_rating.lower() == "outstanding" else "other"
     key += "_10" if close_count > 10 else ""
 
-    series = data[category_name]
+    series = data.loc[(data[category_name] > 0) | (data.index == urn), category_name]
```
PsypherPunk, line 119: suggestion: move this to a distinct function.

I know it's only a line but given what we've been discussing elsewhere, I'd be tempted to pull this into a separate function, something like:

```python
def _get_series_for_category(data: pd.DataFrame, category: str, urn: int) -> pd.Series:
    """
    Some explanation of why we're doing this.
    """
    ...
```

A comment might suffice but it's not so much the _what_ that's complicated, rather the _why_.

WolfyUK (approved):

jrabbott (approved):

PR #1748, opened 2025-01-07
feat: accommodate varying input schemas for Pupil/Workforce Census

PsypherPunk:
### Context

We know that the columns present for each year vary and can require some coaxing into a "standard" format. This is an attempt to move to a flexible way of processing those files.

There's further work to be done: the varies `read_csv`/`read_parquet` calls should be encapsulated elsewhere better but this should give some idea of how we can accommodate these variances.

[AB#242989](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242989)  

### Change proposed in this pull request

- refactor `input_schemas.py` to an input-type specific module
- update processing of pupil-census data to accommodate variance in existing years
- drop `ward_name` column altogether as this is not referenced elsewhere

- update processing of workforce-census data to accommodate variance in  existing years
- extend config. for workforce-census missing from the 2024 data

### Guidance to review 

I've run years 2021 to 2024 through the pipeline locally and all process the input files for their years as expected.

### Checklist (add/remove as appropriate)

- [x] Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)
- [x] Your code builds clean without any errors or warnings
- [x] You have run all unit/integration tests and they pass
- [x] Your branch has been rebased onto main
- [x] You have tested by running locally
- [ ] ~You have reviewed with UX/Design~

PR #1759, opened 2025-01-10
fix: remove null URNs from census data

PsypherPunk:
### Context

Minor follow-up to #1748 to tidy some inconsistencies seen in the output data.

[AB#242989](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242989)  

- explicitly drop any null entries from the URN data before setting the index
- simplify the joining of pupil and workforce data (`join` will default to using the index)

#### `.dropna(subset=["URN"])`

The workforce data contains a `null` row for some of the earlier years. Previously, this was stripped when the `inner` `join` didn't find a match with the pupil data. However, following #1688, the `outer` `join` keeps this in place resulting in an index that is a mix of integers (i.e. URNs) and floats (the `null` is expressed as a float).

#### removal of `on="URN",`

[`join()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.join.html#pandas.DataFrame.join) defaults to using the index: we want to explicitly join in the index and by this point, can guarantee that we want to join both datasets on their respective indexes.

### DIFF
--- a/data-pipeline/src/pipeline/pre_processing.py
+++ b/data-pipeline/src/pipeline/pre_processing.py
@@ -73,12 +73,13 @@ def prepare_census_data(
             na_values=["x", "u", "c", "z", ":"],
             keep_default_na=True,
         )
-        .drop_duplicates()
         .rename(
             columns=input_schemas.workforce_census_column_mappings.get(
                 year, input_schemas.workforce_census_column_mappings["default"]
             ),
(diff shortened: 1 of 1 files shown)


Faizan-ahmad00 (approved):
looks good to me.

PR #1775, opened 2025-01-16
Removed `identifier` dependency when submitting and polling on status of custom school data

WolfyUK:
### Context
[AB#244072](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/244072) [AB#242105](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242105)

- When submitting new/updating existing custom School data, set all matching existing `UserData` rows as `Active = 0` (via earlier modified `POST` Benchmark API endpoint)
- Do not rely on `UserData` identifier when polling for submission status - just check the `Active = 1` row for that user/organisation/type combination (via earlier modified `GET` Benchmark API endpoint)
- If page refreshed during polling stage, do not return to beginning of custom data journey, rather keep polling
- If the user leaves the page after a submission and elects to 'view custom data' when the job is still `pending` then the 'spinner' page should be shown again, rather than a redirect back to self
- E2E tests for Custom Data journey (up to submission page due to otherwise potentially lengthy waits on date pipeline)

### Guidance to review 
Running through the Custom Data journey locally should behave the same after this change as before. The `custom-data` pipeline run may take a long time to complete, so the 'spinner' will poll for a significant period. 

### DIFF
--- a/platform/src/apis/Platform.Api.Benchmark/CustomData/CustomDataService.cs
+++ b/platform/src/apis/Platform.Api.Benchmark/CustomData/CustomDataService.cs
@@ -1,9 +1,8 @@
-using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using Dapper;
 using Dapper.Contrib.Extensions;
-using Platform.Functions.Extensions;
+using Platform.Domain;
 using Platform.Sql;
 
 namespace Platform.Api.Benchmark.CustomData;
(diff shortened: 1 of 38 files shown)


platform/src/apis/Platform.Api.Benchmark/CustomData/CustomDataService.cs
```diff
@@ -12,29 +11,25 @@ public interface ICustomDataService
...
-        _dbFactory = dbFactory;
-    }
-
-
     public async Task UpsertCustomDataAsync(CustomDataSchool data)
     {
         const string sql = "SELECT * from CustomDataSchool where URN = @URN AND Id = @Id ";
 
-        var parameters = new { data.URN, data.Id };
+        var parameters = new
```
jrabbott, line 26: As a team we need to align on formatting settings in our IDE. Would you be able to take this as an action post this PR?

jrabbott (approved):

platform/src/apis/Platform.Api.Benchmark/CustomData/CustomDataService.cs
WolfyUK: Sure. I'll revert this change in the meantime

WolfyUK: I'll create a work item

PR #1815, opened 2025-01-27
docs : Architectural Decision Record (ADR) for Rendering of Charts

jrabbott:
Architectural Decision Record (ADR) for Rendering of Charts


WolfyUK:
Is it worth linking off to any spikes, where applicable?

jrabbott:
> Is it worth linking off to any spikes, where applicable?

@WolfyUK good shout. Would you be able to add a notes section, and add some details based on the spike you'd previously carried out?

WolfyUK (approved):

PR #1884, opened 2025-02-03
feat: make temp. table names unique

PsypherPunk:
### Context

Currently temp. table names are created as `{table_name}_temp`, meaning that should two jobs attempt to write to the same table, one will end up disrupting the other (which can legitimately happen if a `default` and `custom` run happen in parallel, for instance).

[AB#240689](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/240689)  

Firstly, include both the `RunId` value and the current epoch time in the table name (no two jobs occuring at the same time should ever have the same `RunId`).

Secondly, change the Pandas `DataFrame.to_sql()` call to `fail` if there _is_ a collision (this circumstance should never occur and should result in an error).

Thirdly, a refactor of the `pipeline.database` module to avoid the global `engine` variable:

- this was prohibiting unit-testing of this module as it would always attempt to instantiate a SQLAlchemy Engine
- maintain the existing behaviour _by default_ to avoid a large refactor of dependent modules
- allow dependency-injection of an alternative `sqlalchemy.engine.Engine` for testing

Finally, a fix:

- by default, SQLAlchemy/Pandas will make one `INSERT` per row
- [`fast_executemany`](https://docs.sqlalchemy.org/en/20/dialects/mssql.html#fast-executemany-mode) allows multiple rows per `INSERT` _but_ was being incorrectly set at the _cursor_ level, by which point only a single row was being inserted
- set this at the point `create_engine()` is called instead

### Guidance to review 

- I've run the pipeline locally and it writes to the database as previously
- does the refactor for the `engine` seem sensible

**Note**: it's worth checking the "Hide whitespace" option when reviewing as it's clearer that some code has been moved, rather than removed.

### DIFF
--- a/data-pipeline/src/pipeline/database.py
+++ b/data-pipeline/src/pipeline/database.py
@@ -11,33 +11,43 @@ from pipeline import config
 
 logger = logging.getLogger("fbit-data-pipeline")
 
-db_args = os.getenv("DB_ARGS")
 
-args = []
-for sub in db_args.split(";"):
-    if "=" in sub:
-        args.append(map(str.strip, sub.split("=", 1)))
-args = dict(args)
+def get_engine() -> sqlalchemy.engine.Engine:
+    """
+    Creates a SQLAlchemy Engine.
 
-connection_url = URL.create(
-    "mssql+pyodbc",
-    username=os.getenv("DB_USER"),
-    password=os.getenv("DB_PWD"),
-    host=os.getenv("DB_HOST"),
-    database=os.getenv("DB_NAME"),
-    port=os.getenv("DB_PORT"),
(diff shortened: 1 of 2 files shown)


data-pipeline/src/pipeline/database.py
```diff
@@ -11,41 +12,80 @@
...
+            )
+
+    return engine
+
+
+def _get_temp_table(table: str, run_id: str) -> str:
+    """
+    Generate a unique temp. table name.
+
+    The name will comprised of:
```
PsypherPunk, line 63: typo: "…will be comprised of"

J05h-L (approved):

PR #1896, opened 2025-02-06
feat: replace MERGE statements

PsypherPunk:
### Context

Currently we `MERGE` data from a data-pipeline job into each corresponding table. This has led to instances where rows from a previous job have persisted where they should have, in fact, been removed.

[AB#240689](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/240689)  

- replace `MERGE` with `INSERT`/`DELETE` pattern
  - note: for `School`, `Trust`, etc. we not store a per-`RunId` version of the data so they remain as-is for now.
- write _all_ comparator-set/RAG data as a single dataset
  - writing academy/maintained separately means a single `RunId` is spread across two transactions.

### Guidance to review 

~WIP until #1884 is merged; this might need a touch of refactoring afterwards.~

I've run `default`, `custom-data` and `comparator-set` jobs locally and all run as intended.

Noteworthy:

- the temp. table is now defined with a `SELECT…INTO…WHERE 1=0` (which will create an exact copy with no data).
  - previously, there was a mix of specifically defined data-types (see `insert_metric_rag()`) or inferred-from-the-DataFrame (see `insert_financial_data()`).
- the columns required to populate the (temp.) table are retrieved directly from the DB.
  - previously, this was difficult to track due to the various `projections` variables, additional columns added, etc.

### DIFF
--- a/data-pipeline/src/pipeline/database.py
+++ b/data-pipeline/src/pipeline/database.py
@@ -43,6 +43,7 @@ def get_engine() -> sqlalchemy.engine.Engine:
 
     return create_engine(connection_url, fast_executemany=True)
 
+
 @dataclass
 class TempTable:
     name: str
@@ -53,7 +54,10 @@ class TempTable:
         return self.name
 
 
-def _get_table_columns(table: str) -> list[str]:
+def _get_table_columns(
+    table: str,
+    engine: sqlalchemy.engine.Engine,
+) -> list[str]:
     """
     Retrieve the columns for a given table.
 
@@ -64,12 +68,14 @@ def _get_table_columns(table: str) -> list[str]:
     :param table: DB table for which to retrieve columns
     :return: column names
     """
-    sql = textwrap.dedent("""
+    sql = textwrap.dedent(
+        """
     SELECT COLUMN_NAME
       FROM INFORMATION_SCHEMA.COLUMNS
      WHERE TABLE_NAME = :table
     ;
-    """).strip()
+    """
+    ).strip()
 
     with engine.begin() as cnx:
         results = (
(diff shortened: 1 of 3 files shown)


data-pipeline/src/pipeline/database.py
```diff
@@ -29,15 +33,149 @@
...
+    Retrieve the columns for a given table.
+
+    The source-of-truth for the table formats should remain the DB
+    itself: here, we retrieve the list of expected columns directly
+    from the table's schema.
+
+    :param table: DB table for which to retrieve columns
+    :return: column names
+    """
+    sql = textwrap.dedent("""
```
PsypherPunk, line 60: Generally, these changes lean towards deriving as much from the DB as possible: here, for instance, the temp. table is derived from the target table, not defined in the code here.

data-pipeline/src/pipeline/database.py
```diff
@@ -29,15 +33,149 @@
...
+    :param table: target table to which data will be moved
+    :param run_id: unique identifier for processing
+    :return: temp. table details
+    """
+    # TODO: replace when #1884 merged.
+    temp_table_name = f"{table}_{run_id}_temp".replace("-", "_")
+
+    logger.info(f"Creating temp. table: {temp_table_name}.")
+    sql = textwrap.dedent(f"""
+    SELECT *
```
PsypherPunk, line 100: Generally, these changes lean towards deriving as much from the DB as possible: here, for instance, the temp. table is derived from the target table, not defined in the code here.

data-pipeline/src/pipeline/database.py
```diff
@@ -71,69 +209,7 @@ def upsert(
     update_stmt = f'MERGE {table_name} as dest USING {temp_table} as src ON {" AND ".join(match_keys)}  WHEN MATCHED THEN UPDATE SET {", ".join(update_cols)} WHEN NOT MATCHED BY TARGET THEN INSERT ({", ".join(insert_cols)}) VALUES ({", ".join(insert_vals)});'
     with engine.begin() as cnx:
         cnx.execute(sqlalchemy.text(update_stmt))
-        cnx.execute(sqlalchemy.text(f"DROP TABLE IF EXISTS {temp_table}"))
-
-
-def insert_comparator_set(
```
PsypherPunk, line 77: I've just shunted these two functions to the bottom of this module; the function definitions then roughly follow the order in which they're called.

PR #1929, opened 2025-02-11
CSV download of School Census data in a ZIP archive

WolfyUK:
### Context
[AB#249208](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/249208) [AB#242097](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242097)

- CSV download of School Census data via proxy controller, delivered in a ZIP archive
- Designed to honour whatever Pupil comparator set currently displayed for the current user
- Unit test and integration test coverage
- Iterative development included ability to download single CSV rather than via ZIP, which intentionally remains in this PR

### Guidance to review 
To download the ZIP file locally, send a `GET` to (e.g.)  `https://localhost:7095/api/census/download?type=school&id=123456`.

### DIFF
--- a/web/src/Web.App/Controllers/Api/CensusProxyController.cs
+++ b/web/src/Web.App/Controllers/Api/CensusProxyController.cs
@@ -1,11 +1,13 @@
 ﻿using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
+using Web.App.ActionResults;
 using Web.App.Domain;
 using Web.App.Infrastructure.Apis;
 using Web.App.Infrastructure.Apis.Establishment;
 using Web.App.Infrastructure.Apis.Insight;
 using Web.App.Infrastructure.Extensions;
 using Web.App.Services;
+
 namespace Web.App.Controllers.Api;
 
 [ApiController]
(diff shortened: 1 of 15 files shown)


web/src/Web.App/Controllers/Api/CensusProxyController.cs
jrabbott, line 1: [question] why on the proxy controller and not the census controller?

WolfyUK: Happy to move it, but I was following the pattern that seemed to be endpoints called by the client directly are in `/api` and the MVC-style actions are not. This one will be called by a button press on the client, so seemed to be a better fit under `/api`. Also, `/api/census/download` sits alongside `/api/census`, which is the source of the data for the page on which the CTA will preside.

jrabbott: Does this need to be called by the front-end js? Or can it be kept solely within the razor view?

WolfyUK: Not sure yet, as it's a later task. Ideally it would be in the view but I'm not sure how it might affect the styling with the 'save chart images' button (for the School Expenditure version) which _is_ managed via `front-end`.

jrabbott: Ok - it's fine as is. We'll need to be mindful that is might not be the right place for this end-point.

jrabbott:
@J05h-L would you mind pick up the full review?

J05h-L (approved):

PR #1937, opened 2025-02-12
feat: add costs codes to spending priorities and benchmark spending page

J05h-L:
### Context
[AB#242100](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242100) - [AB#248759](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/248759)

Adds CostCodes class to generate the CFR and AAR cost codes and map them to sub category names
A dictionary is passed to the front-end with these cost codes as part of the dataset for the `CompareYourCosts` element for the benchmark spending pages
For spending prorities the sub categories are used to lookup valid values for display.
Adds unit tests and integration tests
Updates selectors on e2e tests following these changes

Note - added for all versions of school benchmark spending - LA, Trust, Custom Data and default School

### Guidance to review 
A bump to front end will follow
E2E tests will be added for benchmark spending following the package update.
Save as Image will include these codes on benchmark spending with the current implementation but more work is required for spending priorities due to the need to exclude certain elements. This will be addressed in a future PR

### DIFF
--- a/front-end-components/src/components/cost-codes-list/component.tsx
+++ b/front-end-components/src/components/cost-codes-list/component.tsx
@@ -1,0 +1,20 @@
+import { CostCodesListProps } from "src/components/cost-codes-list";
+import { useCostCodeMapContext } from "src/contexts/hooks";
+
+export const CostCodesList: React.FC<CostCodesListProps> = (props) => {
+  const { category } = props;
+
+  const { categoryCostCodes } = useCostCodeMapContext(category);
+
+  return (
+    categoryCostCodes && (
+      <ul className="app-cost-code-list">
+        {categoryCostCodes.map((costCode) => (
+          <li key={costCode}>
+            <strong className="govuk-tag">{costCode}</strong>
+          </li>
+        ))}
+      </ul>
+    )
+  );
+};
--- a/front-end-components/src/contexts/hooks.tsx
+++ b/front-end-components/src/contexts/hooks.tsx
@@ -4,6 +4,8 @@ import {
   ChartModeContextValue,
   CentralServicesBreakdownContext,
   CentralServicesBreakdownContextValue,
+  CostCodeMapContext,
+  CostCodeMapContextValues,
 } from "./contexts";
 
 export const useChartModeContext = (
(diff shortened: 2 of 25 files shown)


web/src/Web.App/AssetSrc/scss/main.scss
```diff
@@ -413,4 +413,11 @@ hr.govuk-section-break--print {
 
 .composed-container {
   min-height: 335px;
-}
\ No newline at end of file
+}
+.app-cost-code-list {
```
WolfyUK, line 417: I'm assuming this is here rather than in `front-end` because it is consumed by `Web` also?

J05h-L: Correct yes, needed for the spending priorities page also where this is outside of the chart component.

front-end-components/src/components/cost-codes-list/component.tsx
```diff
@@ -0,0 +1,222 @@
+import { useContext } from "react";
+import { CostCodesListProps } from "src/components/cost-codes-list";
+import { CostCodeMapContext } from "src/contexts";
+
+export const CostCodesList: React.FC<CostCodesListProps> = (props) => {
+  const { category } = props;
+
+  const costCodeMap = useContext(CostCodeMapContext);
+
+  const getCostCodes = (
```
WolfyUK, line 10: To prevent duplication of the category-to-code(s) logic, would it be possible to change the `costCodeMap` sent to `front-end` from a `Dictionary<string, string>` to a `Dictionary<string, string[]>`, where the key is the category and the values are the codes? It looks like this wouldn't be an insignificant amount of work, but I'm thinking of what happens when new sub-categories come along or are re-parented.

front-end-components/src/components/cost-codes-list/component.tsx
```diff
@@ -0,0 +1,222 @@
...
+          ],
+        ];
+      }
+      default: {
+        return [];
+      }
+    }
+  };
+
+  const costCodes = getCostCodes(category, costCodeMap).filter((code) => code);
```
WolfyUK, line 209: What is the `filter` doing here? If it is a truthy check, then it might be better to do `filter((code) => !!code)

J05h-L: yes truthy check, will update

front-end-components/src/components/cost-codes-list/component.tsx
WolfyUK: Ah, I see. How irritating. This goes back to our conversation around chart IDs I guess. Maybe leave as-is for now to get this in, but it would be good to get a follow-up work item to look at again.

WolfyUK: As per discussion, leave logic as-is for the time being, but move `useContext(CostCodeMapContext)` into a new hook in `/src/contexts/hooks.tsx` that returns the `getCostCodes()` function directly, keeping the internals out of the `<CostCodesList />` component.

front-end-components/src/components/cost-codes-list/component.tsx
```diff
@@ -0,0 +1,22 @@
+import { CostCodesListProps } from "src/components/cost-codes-list";
+import { useCostCodeMapContext } from "src/contexts/hooks";
+
+export const CostCodesList: React.FC<CostCodesListProps> = (props) => {
+  const { category } = props;
+
+  const { getCostCodes } = useCostCodeMapContext();
```
WolfyUK, line 7: Sorry, one for a future refactor, but the new hook could also be extended to accept `category` as an argument and then return the `costCodes` as `categoryCostCodes` or similar, alongside 'all' cost codes. This would be instead of the `getCostCodes()` function (i.e. line 9 from here moves inside the hook). Happy for this to be done at a later date.

front-end-components/src/contexts/hooks.tsx
```diff
@@ -39,3 +41,205 @@ export const useCentralServicesBreakdownContext = (
...
 };
+
+export const useCostCodeMapContext = (): CostCodeMapContextValue => {
+  const costCodeMap = useContext(CostCodeMapContext);
+
+  const getCostCodes = (category: string): string[] => {
+    if (!costCodeMap) {
+      return [];
+    }
+    switch (category) {
```
WolfyUK, line 52: Maybe add a comment here to advise that this mapping knowingly duplicates what is done in Web

WolfyUK (approved):

front-end-components/src/components/cost-codes-list/component.tsx
J05h-L: happy to do that now, you're right that's worthwhile and a quick change

J05h-L: Updated as suggested

WolfyUK (approved):

PR #1947, opened 2025-02-13
feat: include cost code list in saved image for spending priorities

J05h-L:
### Context
[AB#242100](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/242100) - [AB#249508](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/249508)

Similar to how we create the chart titles for the spending priorities page the we pass the cost codes through the `html-dataset` so that they can be used in `onCloned` to create the required elements for the image and set styles.
Slight refactor around `onCloned` as it is too busy with the addition of creating cost codes as well but previous behaviour should be preserved.

### Guidance to review 
Is there a better way to achieve this?

### DIFF
--- a/front-end-components/src/components/share-content/types.tsx
+++ b/front-end-components/src/components/share-content/types.tsx
@@ -14,4 +14,5 @@ export type ShareContentProps = Pick<
   saveEventId?: string;
   showCopy?: boolean;
   showSave?: boolean;
+  costCodes?: string[];
 };
--- a/front-end-components/src/hooks/useDownloadImage.ts
+++ b/front-end-components/src/hooks/useDownloadImage.ts
@@ -11,26 +11,30 @@ export type DownloadPngImageOptions<T> = {
   onSaved?: (fileName: string) => void;
   ref?: React.RefObject<T>;
   showTitle?: boolean;
+  costCodes?: string[];
   title?: string;
 } & Pick<ImageOptions, "filter">;
 
-export type ElementAndTitle = {
+export type ElementAndAttributes = {
   element: HTMLElement;
   title?: string;
+  costCodes?: string[] | undefined;
 };
 
 export type DownloadPngImagesOptions = {
   fileName?: string;
-  elementsSelector: () => ElementAndTitle[];
+  elementsSelector: () => ElementAndAttributes[];
   onImagesLoading?: (loading: boolean) => void;
   onProgress?: (percentage: number) => void;
   showTitles?: boolean;
+  costCodes?: string[];
   signal?: AbortSignal;
 } & Pick<ImageOptions, "filter">;
 
 const imageTitleHeight = 50;
 const type = "image/png";
 const backgroundColor = "#fff";
+const CostCodesListHeight = 50;
 
 export function useDownloadPngImage<T>({
   elementSelector,
@@ -40,6 +44,7 @@ export function useDownloadPngImage<T>({
   onLoading,
   onSaved,
   showTitle,
+  costCodes,
   ref,
   title,
 }: DownloadPngImageOptions<T>) {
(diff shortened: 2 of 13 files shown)


front-end-components/src/components/share-content/types.tsx
```diff
@@ -14,4 +14,5 @@ export type ShareContentProps = Pick<
   saveEventId?: string;
   showCopy?: boolean;
   showSave?: boolean;
+  costCodes?: string[] | undefined;
```
WolfyUK, line 17: `| undefined` is pretty much redundant if `codeCodes` is also nullable

front-end-components/src/hooks/useDownloadImage.ts
```diff
@@ -244,19 +251,22 @@ const getImageOptions = (
...
-      child.style.height = `${imageTitleHeight}px`;
-      child.style.margin = `${imageTitleHeight / 4}px 0 -${imageTitleHeight / 4}px 0`;
-      child.style.padding = "0";
-      node.insertBefore(child, node.firstChild);
+  }
+  if (costCodes) {
+    height += CostCodesListHeight;
+  }
+
+  if ((title && showTitle) || costCodes) {
```
WolfyUK, line 262: Almost redundant conditional - this check is done again within the `onCloned()` function, so probably isn't needed twice.

web/src/Web.App/Views/SchoolSpending/Index.cshtml
```diff
@@ -11,7 +11,9 @@
         saveButtonVisible = true,
         saveClassName = "costs-chart-wrapper",
         saveFileName = $"spending-priorities-{Model.Urn}.zip",
-        saveTitleAttr = "data-title"
+        saveTitleAttribute = "data-title",
+        saveTitleAttr = "data-title",
+        CostCodesAttr = "data-cost-codes"
```
WolfyUK, line 16: Should be camel case on `CostCodesAttr`

WolfyUK:
> Is there a better way to achieve this?

No, this looks great!

front-end-components/src/hooks/useDownloadImage.ts
J05h-L: this was to leave `onCloned` as it currently is and only assign it when needed. Wasn't sure if preserving the behaviour for leaving it sometimes `undefined` was better?

front-end-components/src/components/share-content/types.tsx
J05h-L: thanks will update

web/src/Web.App/Views/SchoolSpending/Index.cshtml
J05h-L: thanks! will update

front-end-components/src/hooks/useDownloadImage.ts
WolfyUK: I see. For readability, I'd be happier if it just did nothing in the case when there is neither a title nor tags to build, rather than be `undefined`

J05h-L: Understood - yeah that makes sense - will update it also now, thanks!

front-end-components/src/hooks/useDownloadImage.ts
```diff
@@ -11,26 +11,30 @@ export type DownloadPngImageOptions<T> = {
...
   onProgress?: (percentage: number) => void;
   showTitles?: boolean;
+  costCodes?: string[];
   signal?: AbortSignal;
 } & Pick<ImageOptions, "filter">;
 
 const imageTitleHeight = 50;
 const type = "image/png";
 const backgroundColor = "#fff";
+const CostCodesListHeight = 50;
```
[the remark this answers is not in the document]
J05h-L: my bad! Fixing now

WolfyUK (approved):

PR #2031, opened 2025-02-26
fix: use chunksize for df.to_sql

PsypherPunk:
### Context

Writing of `MetricRAG` to the temp. table is taking 60 minutes+.

[AB#250420](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/250420)  

Use the [`chunksize` parameter](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html) of Pandas' `to_sql()` method to reduce the batch size (configurable by an environment variable but with a default of 10,000).

### Guidance to review 

Verified locally that data are still correctly written to the DB.

### DIFF
--- a/data-pipeline/src/pipeline/database.py
+++ b/data-pipeline/src/pipeline/database.py
@@ -192,6 +192,7 @@ def _write_data(
         con=engine,
         if_exists="append",
         index=df.index.name is not None,
+        chunksize=int(os.getenv("SQL_CHUNKSIZE", 10_000)),
     )
     logger.info(
         f"Wrote {len(df.index):,} rows to {temp_table} in {int(time.time() - start):,} seconds."


data-pipeline/src/pipeline/database.py
```diff
@@ -192,6 +192,7 @@ def _write_data(
         con=engine,
         if_exists="append",
         index=df.index.name is not None,
+        chunksize=int(os.getenv("SQL_CHUNKSIZE", 10_000)),
```
WolfyUK, line 195: Is `SQL_CHUNKSIZE` defined somewhere already?

PsypherPunk: It's reading from an environment variable; if it's set, it'll pick up that value, if not it'll default to 10,000.

I'd originally hard-coded 10,000 but added this in case we want to optimise in the future.

WolfyUK: Ah, OK. Wasn't sure if you wanted to support some custom value via env var from the outset

PR #2076, opened 2025-03-04
Management and persistence of comparator LAs for High needs benchmarking

WolfyUK:
### Context
[AB#252038](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/252038) [AB#249552](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/249552)

- Select/add/remove LAs (via `<select>` only in current iteration)
- Validation for selection and min/max items
- Save to session or cancel
- Integration tests

### Guidance to review 
Deployed to `d18` feature environment for validation. Progressive enhancement to follow as part of future work. Copy review as part of future work (especially around validation messages).

### DIFF
--- a/web/Web.sln.DotSettings
+++ b/web/Web.sln.DotSettings
@@ -38,6 +38,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=gias/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=govuk/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=icfp/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=nfer/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=noopener/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=noreferrer/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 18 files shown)


jrabbott (commented):
I'm unclear on some of the choices on implementation details. Could we chat them through?

For example; I'm unclear why session state is being used, as well as the import/export of model state.

jrabbott:
If we look at `OtherTeachingPeriods` on the ICFP journey which mirrors the `add another thing` pattern. This has been implemented without the need for import/export model state, nor the need for additional query parameters or session state.

Unless I'm missing something I feel we should be able to have a simple implementation of this where the addition and removal of LA's from the comparator set is handle within the standard model state. We'll need to ensure the list of comparators is populated, however this can be handled with an API call as part of the post request (if remove or add).

WolfyUK:
Sure, @jrabbott. Happy to catch up today, as I was expecting to iterate on this first version.

> why session state is being used

I believe this was part of the original requirements for the first iteration. Once the user has chosen their comparators (by clicking 'Save and continue') they are saved in Session state only, rather than persisted to the database.

> import/export of model state

This is to implement the PRG pattern, although model state should be used rather than query strings (which will be fixed now). Currently the export/import model state only contains validation errors.

> `add another thing` pattern

I am happy to re-write to mirror ICFP, but this seems like a much more complex problem than what we are trying to resolve here, and from what I can tell from the [DWP](https://design-system.dwp.gov.uk/patterns/add-another-thing), [HMRC](https://design.tax.service.gov.uk/hmrc-design-patterns/add-to-a-list/) and [Home Office](https://design.homeoffice.gov.uk/patterns/add-multiple-things) examples, this is a multi-page pattern, where we want everything on the same page.

jrabbott (approved):

PR #2088, opened 2025-03-07
feat: Local Authority statistical neighbours data

J05h-L:
### Context
[AB#250505](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/250505) - [AB#252266](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/252266)

- Update data-pipeline to process the statistical neigbours data with the required columns names
- Combine the above with the existing section 251 dataframe to make a local authority dataframe which is used to create a parquet (and later to write to the db)
- Test coverage
- Update data sources docs

### Guidance to review 
Is the approach suitable?
pipeline should run successfully locally and the `local_authorities.parquet` should have the correct columns and values as detailed in the ticket in addition to the existing S251 columns


data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
```diff
@@ -25,8 +29,20 @@ def build_local_authorities(
...
 
-    # TODO: combined with out LA datasets.
-    return section_251_data
+    logger.info("Processing Local Authority statistical neighbours data.")
+
+    la_statistical_neighbours_data = _prepare_la_statistical_neighbours(
+        statistical_neighbours_filepath, year
+    )
+
+    local_authority_data = section_251_data.merge(
```
J05h-L, line 38: [Question] We merge with the existing section 251 data. Some records may be present in the `la_statistical_neighbours_data` and not in `section_251_data`. Expect this will be LA that have split.

This will lose the SN data for those authorities. Here I have assumed that is desired. Is it?

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
```diff
@@ -9,13 +9,17 @@
...
+        FilePath | ReadCsvBuffer[bytes] | ReadCsvBuffer[str]
+    ),
     year: int,
 ):
     """
     Build Local Authority data from various sources.
 
     :param budget_filepath_or_buffer: source for LA budget data
     :param outturn_filepath_or_buffer: source for LA outturn data
+    :param statistical_neighbours_filepath: source for Local Authority data
```
PsypherPunk, line 22: nitpick: `source for Local Authority statistical neighbours data`…?

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
PsypherPunk: Tagging @benlav-50 on this: what's the "source of truth" for which LAs should be in the final dataset?

I _think_ we'd said that S251 was the "main" dataset but good to confirm.

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
```diff
@@ -25,8 +29,20 @@ def build_local_authorities(
...
+
+    la_statistical_neighbours_data = _prepare_la_statistical_neighbours(
+        statistical_neighbours_filepath, year
+    )
+
+    local_authority_data = section_251_data.merge(
+        la_statistical_neighbours_data,
+        left_on="old_la_code",
+        right_index=True,
+        how="left",
```
PsypherPunk, line 42: question: does the `section_251_data` need a `reset_index()` before doing this? IIRC, it's got a multi-index of `old_la_code` and `new_la_code` which won't match…?

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
```diff
@@ -213,3 +229,37 @@ def _prepare_la_section_251_data(
...
+        usecols=input_schemas.la_statistical_neighbours.get(
+            year, input_schemas.la_statistical_neighbours["default"]
+        ).keys(),
+    ).rename(
+        columns=input_schemas.la_statistical_neighbours_column_mappings.get(
+            year, input_schemas.la_statistical_neighbours_column_mappings["default"]
+        ),
+    )
+
+    df.dropna(how="all", inplace=True)
```
PsypherPunk, line 263: question: just to check that this is _definitely_ how we want to drop any null values?

IIRC the S251 data, for instance, has at least one column with entirely nulls which would be removed but we don't want that—could/should this be limited to specific columns?

PsypherPunk (commented):
Couple of minor queries but this looks good.

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
PsypherPunk: No, all good: if Pandas is behaving itself (and I'll admit _that's_ the surprising part) it's good as it is.

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
PsypherPunk: FWIW, something like `df = df[~df.index.isna()]` happens elsewhere (and sounds like it would be relevant here).

J05h-L: I was thinking of

```
    ).dropna(
        subset=[input_schemas.la_statistical_neighbours_index_col]
    ).set_index(
        input_schemas.la_statistical_neighbours_index_col
    )
```

but your suggestion is better, I will go with that

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
J05h-L: updated with a better description

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
J05h-L: Updated as suggested

data-pipeline/src/pipeline/pre_processing/ancillary/local_authority.py
PsypherPunk: As per conversation in Slack: the S251 data can be considered the "main" dataset so this is correct in being a left-join (tagging @jrabbott for visibility).

jrabbott: Reasoning being; organisations (schools/trusts/la) must submit financial data. Therefore no financial data, no organisation.

PR #2153, opened 2025-03-25
feat: add dict variant for multiple-keys

PsypherPunk:
### Context

Now that we define variable input-schemas for each year, for a given data source, there are numerous occurrences where the same year will have the same config. (but not in the current/majority so cannot be considered a "default" value).

Python doesn't have a data structure to accommodate this, hence this addition.

add a `dict` variant to accommodate multiple keys addressing the same value: this should a fair amount of repetition in some of the `input_schemas` definitions.

### Guidance to review 

WIP for discussion.

- [ ] ~Work items have been linked [(use AB#)](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops#use-ab-to-link-from-github-to-azure-boards-work-items)~

### DIFF
--- a/data-pipeline/src/pipeline/utils/input.py
+++ b/data-pipeline/src/pipeline/utils/input.py
@@ -1,0 +1,48 @@
+from collections import abc
+
+
+class MultiKeyDefaultDict(abc.Mapping):
+    """
+    Variant of `dict` which should:
+
+    - allow multiple keys to mapped to the same value
+    - always return a (mandatory) default if a key is missing
+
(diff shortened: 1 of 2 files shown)


J05h-L (commented):
I think this will be helpful. I think the implementation will be easy enough to use/understand for us and future maintainers and definitely helps tidy up some of the input schemas

Faizan-ahmad00 (commented):
Looks good to me from maintenance POV.

Faizan-ahmad00 (approved):

J05h-L (approved):

PR #2155, opened 2025-03-26
feat: create new index to support school faceted search WIP

J05h-L:
### Context
AB#250281 / AB#255008

Creates new index and indexer for `SchoolFacetedIndex` using the existing `School` data source
Currently the same fields as the `SchoolIndex` with the addition of `OverallPhase` and `SchoolType`

### Guidance to review 
WIP


WolfyUK (approved):

jrabbott:
@J05h-L / @WolfyUK - just catching up so I'm probably missing something, what's the purpose of the new index (i.e. why couldn't the school index be used?)

J05h-L:
@jrabbott @WolfyUK rather than changing the existing index, `SimpleField` to `SearchableField` on a number of fields and the addition of new fields we felt it was better to create a new index using the same source. 

Happy to change this though if you think amending the existing school index is better?

jrabbott:
@J05h-L thanks - I'm still unclear on the why? and what the purpose would be of the two indexes (`SchoolFacetedIndex` and `SchoolIndex`)

At the moment, I'm not sure which approach would be "better".

J05h-L:
@jrabbott the purpose of two was to keep them separate so the addition of the new fields and the changes required to the existing fields wouldn't impact the behaviour `SchoolIndex`. There is a lot of overlap however...

PR #2160, opened 2025-04-01
School search results view/filter/paginate

WolfyUK:
### Context
[AB#255010](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/255010) [AB#255011](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/255011) [AB#255012](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/255012) [AB#255425](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/255425) [AB#250281](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/250281)

- Search results page, with shared search term form moved to partial
- Explicit App Insights feature request tracking
- Pagination view component
- Populated search results view
- Sort and filtering support
- Added search results list to results page
- Integration tests for Search Results page (this will require another pass once API added)

image

### Guidance to review 
A set of static results are hard coded in this PR. Wiring up to the API will take place as part of other work items. It is also anticipated that an amount of refactoring will take place as part of other user stories when it comes to implementing faceted search for other establishment types.


web/src/Web.App/ViewModels/Components/PaginationViewModel.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            return pages.ToArray();
+        }
+    }
+    public bool SkipBeforeMidPages => MidPages.FirstOrDefault(FirstPage) > FirstPage + 1;
+    public bool SkipAfterMidPages => MidPages.LastOrDefault(LastPage) < LastPage - 1;
+    public bool HasPreviousPage => CurrentPage > FirstPage;
+    public bool HasNextPage => CurrentPage * pageSize < totalResults;
+
+    public string BuildPageUrl(int page)
```
jrabbott, line 32: Why not utilise `Url.Action` for building the URL?

web/src/Web.App/Views/SchoolSearch/Index.cshtml
jrabbott, line 1: [question] - what is the expected behaviour if I change the search term but press apply? equally if I change the filter/sort but press search?

Feels like there are two forms on this page that should be independent of each other (i.e. apply ignores changes to search terms, and search ignores filter/sort if new term is entered)

web/src/Web.App/ViewModels/Components/PaginationViewModel.cs
WolfyUK: It was the intention that this component be generic enough to be reusable elsewhere. Its only responsibility should be the `page` query string value, and all the others should be left untouched.

web/src/Web.App/Views/SchoolSearch/Index.cshtml
WolfyUK: > if I change the search term but press apply

The new search term will be used, plus the current filter selection (if any present)

>  if I change the filter/sort but press search

The existing search term will be used with the new filters.

You are correct that the single form causes this behaviour. Happy to extract out into two forms again, as per an earlier iteration. This would also move the 'search' form back into the SearchTerm partial and add a hidden `<input>` with the original search term into a new 'filter' form.

web/src/Web.App/ViewModels/Components/PaginationViewModel.cs
jrabbott: Still not sure why Url.Action couldn't be used for building the url? Happy to jump on a call to discuss

web/src/Web.App/Views/SchoolSearch/Index.cshtml
jrabbott: It probably needs the interaction/behaviour confirming - I'm just going off my expectation.

WolfyUK:
> discussed in refinement that Education Phase should be checkboxes

I do not recall this, and nothing is mentioned on the story. How is it expected this would work? Multiple facet selection via check boxes does not make sense here. Do you mean exclusive checkboxes? Is that permitted under GDS?

web/src/Web.App/Controllers/FindOrganisationController.cs
```diff
@@ -110,22 +111,27 @@ private IActionResult FacetedSearchResult(FindOrganisationSelectViewModel viewMo
...
     {
-        return View();
+        return View(new FindSchoolViewModel());
     }
 
     [HttpPost]
     [Route(OrganisationTypes.School)]
     [FeatureGate(FeatureFlags.FacetedSearch)]
-    public IActionResult School(FindOrganisationSearchViewModel viewModel)
+    public IActionResult School(FindSchoolViewModel viewModel)
```
jrabbott, line 123: Feels like this and the above methods/routes should be part of the school search controller?

web/src/Web.App/ViewModels/Components/PaginationViewModel.cs
WolfyUK: Yes, an example might be handy here

web/src/Web.App/Views/SchoolSearch/Index.cshtml
WolfyUK: I think this is a valid expectation, although it is not explicitly specified on the story

web/src/Web.App/Controllers/FindOrganisationController.cs
```diff
@@ -110,22 +111,27 @@ private IActionResult FacetedSearchResult(FindOrganisationSelectViewModel viewMo
...
         {
-            return View();
+            return View(viewModel);
         }
 
-        return NotFound();
+        return RedirectToAction("Index", "SchoolSearch", new
+        {
+            term = viewModel.Term,
+            redirect = bool.TrueString.ToLower()
```
jrabbott, line 133: Unclear on the redirect prop?

jrabbott:
> I do not recall this, and nothing is mentioned on the story. How is it expected this would work? Multiple facet selection via check boxes does not make sense here. Do you mean exclusive checkboxes? Is that permitted under GDS?

Think it's subjective - using GIAS as example, I'd expect when search for I school to be able to select one or more phases if I'm not sure what phase the school is.

image

jrabbott:
> Think it's subjective - using GIAS as example, I'd expect when search for I school to be able to select one or more phases if I'm not sure what phase the school is.


Checkboxes would be `or` not `and`

web/src/Web.App/Controllers/FindOrganisationController.cs
jrabbott: Could we do now so we don't end up with rework on the school journey, when we add the others.

web/src/Web.App/Views/SchoolSearch/Index.cshtml
jrabbott: Yeah, worth reaching out to designers and confirming expectation

web/src/Web.App/Controllers/FindOrganisationController.cs
WolfyUK: Updated ✅

web/src/Web.App/Controllers/FindOrganisationController.cs
WolfyUK: Updated ✅

WolfyUK:
> > I do not recall this, and nothing is mentioned on the story. How is it expected this would work? Multiple facet selection via check boxes does not make sense here. Do you mean exclusive checkboxes? Is that permitted under GDS?
> 
> Think it's subjective - using GIAS as example, I'd expect when search for I school to be able to select one or more phases if I'm not sure what phase the school is.

Updated ✅

jrabbott (approved):

PR #2223, opened 2025-04-03
feat: updates to `SchoolIndex` to support faceted school search

J05h-L:
### Context
AB#250281 - AB#255218 - AB#256038

following conversations on #2155 A seperate index is not required to support faceted search, instead use the existing `SchoolIndex` amended where required.

- Amends `SchoolIndex` to contain searchable fields and added an additional facetable field for `OverallPhase`
- Reverts the changes introduced on #2155

### Guidance to review 
deploying to d15

- [ ] You have reviewed with UX/Design


WolfyUK:
@J05h-L Does re-using the index allow searches to be done against different fields? For example, the suggester searches all searchable fields but the faceted search only searches 'school name'?

J05h-L:
We've had confirmation now that the search should work the same as the suggester in that we can search by URN, name, and address fields. ACs have been updated

jrabbott:
@J05h-L / @WolfyUK I think there is some confusion here on suggest and search, and how the index configuration is setup to support both of these.

J05h-L:
@jrabbott / @WolfyUK my understanding is that the suggester can use any of the fields they just need to be added to `options.Select` when we create the `SuggestOptions`

For search the field needs to be marked as `SearchableField`

Please correct me if I am misunderstanding something here.

jrabbott:
On an index we could have multiple suggesters, each having their own configuration for which fields to suggestion on (these do not have to be 'searchable' fields - see previous configuration, none of the fields where searchable).

'Suggest' is a different operation to 'Search' and as such we have to specify both the index and suggester to use. Options on the suggest request doesn't change the configuration of the index (i.e which fields to suggest on) rather it defines which fields to return in the response.

'Searchable' attributes relate to the 'Search' operation. 

All fields in a given suggester configuration will be used in suggest operation.

All fields marked as searchable will be used in search operation.

WolfyUK (approved):

PR #2339, opened 2025-05-07
Applied new `[ValidateLaCode]` attribute to actions and controllers that consume `code`

WolfyUK:
### Context
[AB#254608](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/254608) [AB#260043](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/260043) #2337 

- Implemented action filter above for LA codeand applied to controllers/actions
- Updated integration tests to ensure all dummy LA codes are of correct format
- Refactored earlier implementation to instead make use of `IValidator`

### Guidance to review 
Dependency on #2337.
Any URL including a bad LA code should fail with `404` and the 'page not found' view. E2E/A11y should be unaffected.

### DIFF
--- a/web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
+++ b/web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
@@ -23,21 +23,24 @@ public class OrganisationIdentifierValidator : AbstractValidator<OrganisationIde
 
         RuleFor(p => p.Value)
             .Length(6)
+            .Must(p => p.All(char.IsDigit))
             .When(p => p.Type == OrganisationTypes.School)
             .When(p => p.ValueAsInt != null)
-            .WithMessage("School URN must be 6 characters");
+            .Configure(rule => rule.MessageBuilder = _ => "School URN must be 6 digits");
 
         RuleFor(p => p.Value)
             .Length(8)
+            .Must(p => p.All(char.IsDigit))
             .When(p => p.Type == OrganisationTypes.Trust)
             .When(p => p.ValueAsInt != null)
-            .WithMessage("Trust company number must be 8 characters");
+            .Configure(rule => rule.MessageBuilder = _ => "Trust company number must be 8 digits");
 
         RuleFor(p => p.Value)
             .Length(3)
+            .Must(p => p.All(char.IsDigit))
             .When(p => p.Type == OrganisationTypes.LocalAuthority)
             .When(p => p.ValueAsInt != null)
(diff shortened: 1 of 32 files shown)


web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        RuleFor(p => p.Type)
+            .Must(type => _organisationTypes.Contains(type))
+            .When(p => p.Type != null)
+            .WithMessage("Organisation type must be valid");
+
+        RuleFor(p => p.Value)
+            .Length(6)
+            .When(p => p.Type == OrganisationTypes.School)
+            .When(p => p.ValueAsInt != null)
```
jrabbott, line 27: Would -19823 pass this validation as is? Length 6 and is a valid int?

Could use `.Must(p=> p.All(char.IsDigit))`?

web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
jrabbott, line 1: Have you seen the rule set feature in FluentValidation? Not sure if it'll help make the logic in the validator easier to follow rather than `.When(p => p.Type == OrganisationTypes.School)`.

Not suggesting we have to use rule sets.

https://docs.fluentvalidation.net/en/latest/rulesets.html

WolfyUK: Yes, I read that earlier but failed to see how it would help in this case

web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
WolfyUK: Good spot! Updated

J05h-L (approved):

web/src/Web.App/Validators/OrganisationIdentifierValidator.cs
jrabbott: What about XXXXXX? Does that pass or fail the validation?

My gut is saying that'll pass based on the 'when'. However I've not used when before so not 100% sure on the behaviour.

WolfyUK: One of the other rules fails validation with `Organisation identifier must be a number` for `XXXXXX`

PR #2341, opened 2025-05-07
Applied new `[ValidateId]` attribute to proxy controller actions that consume `id` and `type`

WolfyUK:
### Context
[AB#254608](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/254608) [AB#260043](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/260043) #2339 

- Implemented action filter above for School/Trust/LA identifiers and applied to proxy controllers actions
- Updated integration tests to ensure all dummy identifiers are of correct format

### Guidance to review 
Dependency on #2339.
Any URL including a bad URN/Company Number/LA code for their respective organisation type should fail with `404` and the 'page not found' view. E2E/A11y should be unaffected.

### DIFF
--- a/web/src/Web.App/Attributes/ValidateArgumentAttribute.cs
+++ b/web/src/Web.App/Attributes/ValidateArgumentAttribute.cs
@@ -87,8 +87,10 @@ internal class ValidateArgumentFilter(
     }
 }
 
-public class ValidateUrnAttribute() : ValidateArgumentAttribute("urn", OrganisationTypes.School);
+public class ValidateUrnAttribute(string argumentName = "urn") : ValidateArgumentAttribute(argumentName, OrganisationTypes.School);
 
-public class ValidateCompanyNumberAttribute() : ValidateArgumentAttribute("companyNumber", OrganisationTypes.Trust);
+public class ValidateCompanyNumberAttribute(string argumentName = "companyNumber") : ValidateArgumentAttribute(argumentName, OrganisationTypes.Trust);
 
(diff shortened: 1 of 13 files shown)


jrabbott:
A couple of things I’m unclear on:

- I noticed the validation approach doesn’t use DataAnnotations or FluentValidation - just curious about the reasoning behind that. Could be a simpler way to handle it?
- The base and specific Id filter doesn’t fully follow the Open/Closed Principle (OCP). It feels like we might need to update the base class every time a new filter is added. Maybe there’s a way to make it more extendable and maintainable?
- Also, the view response doesn’t quite match with the status code being returned, especially when handling different responses for Controller (View) and Api Controller (JSON). Might be worth a quick chat to clarify how this is supposed to work.

If you’ve got 5 mins, it might be easier to discuss these on a call.

WolfyUK:
> * I noticed the validation approach doesn’t use DataAnnotations or FluentValidation - just curious about the reasoning behind that. Could be a simpler way to handle it?
> * The base and specific Id filter doesn’t fully follow the Open/Closed Principle (OCP). It feels like we might need to update the base class every time a new filter is added. Maybe there’s a way to make it more extendable and maintainable?
> * Also, the view response doesn’t quite match with the status code being returned, especially when handling different responses for Controller (View) and Api Controller (JSON). Might be worth a quick chat to clarify how this is supposed to work.

Resolved in dependent PR and rebased here

PR #2345, opened 2025-05-08
Defined Front Door redirect rule 

WolfyUK:
### Context
[AB#260470](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/260470) [AB#260471](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/260471)

Created Front Door redirect rule to ensure all traffic to non-primary host names is sent to the root path of the primary host name.

### Guidance to review 
Deployed to `d11` feature environment to validate behaviour. Additional endpoint manually associated with Front Door profile in order to test non-primary host name redirection logic.

### DIFF
--- a/web/terraform/README.md
+++ b/web/terraform/README.md
@@ -32,6 +32,8 @@ No modules.
 | [azurerm_cdn_frontdoor_origin_group.web-app-front-door-origin-group](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_origin_group) | resource |
 | [azurerm_cdn_frontdoor_profile.web-app-front-door-profile](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_profile) | resource |
 | [azurerm_cdn_frontdoor_route.web-app-front-door-route](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_route) | resource |
+| [azurerm_cdn_frontdoor_rule.web-app-redirect-rule](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/cdn_frontdoor_rule) | resource |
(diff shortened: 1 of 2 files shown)


J05h-L (approved):
lgtm

jrabbott:
Browsers can/do aggressively cache the permanent redirect. This will impact/impose future constraints

It isn't clear from the ticket why are doing this / what problem we are solving with this change?

WolfyUK:
> Browsers can/do aggressively cache the permanent redirect. This will impact/impose future constraints

Happy to switch to a `307 Temporary redirect`.

> It isn't clear from the ticket why are doing this / what problem we are solving with this change?

Discussions going on throughout yesterday and today have led to this requirement. It was also listed on the [original decision](https://github.com/DFE-Digital/education-benchmarking-and-insights/blob/main/documentation/architecture/decisions/0010-sfb-decomm.md#option-2-dns-level-redirect-to-the-new-service) but implementation was omitted as part of the decommissioning work intended to be deployed last release.

jrabbott:
I think there is some confusion. The document doesn't describe providing a redirect response from our front door.

Without full context (which is missing from the user story) 307 should be used to avoid caching issues.

WolfyUK:
> Without full context (which is missing from the user story) 307 should be used to avoid caching issues.

Switched to `307 Temporary redirect` and added more detail to the work item.

jrabbott:
Keep in mind the pervious implementation was intentional.

If the user navigated to an unsupported URL/route they would see a 404 error page and be able to take the necessary action.

With this implementation the user won't have this experience, and will be silently taken to the service home page without knowing why.

WolfyUK:
> With this implementation the user won't have this experience, and will be silently taken to the service home page without knowing why.

Spot on. This has been agreed with the client.

jrabbott:
Decision needs to be documented as this is a poor user experience.

Faizan-ahmad00 (approved):

PR #2400, opened 2025-05-23
Return `Ssr` suffixed views if feature enabled

WolfyUK:
### Context
[AB#258039](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/258039) [AB#261663](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/261663)

- Return `Ssr` suffixed views if feature enabled, rather than determine in view components or partials
- This will allow the non-`Ssr` views to be deleted once the feature is enabled and rolled out rather than having to hand pick through code to delete
- Added warning message if SSR enabled but charts could not be loaded
- Updated integration tests

### Guidance to review 
To test the warning message scenario make sure the SSR feature is enabled and then either do not run the `chart-rendering` function app if running locally, or change `appsettings.json` to point to a dummy endpoint. Polly will attempt to resolve a number of times, so the response may be slow. A later consideration may be to use an alternative Polly policy when hitting the charting API.

### DIFF
--- a/web/src/Web.App/Views/Shared/Components/PageActions/Default.cshtml
+++ b/web/src/Web.App/Views/Shared/Components/PageActions/Default.cshtml
@@ -4,35 +4,28 @@
     <div class="page-actions">
         @if (Model.SaveButtonVisible)
         {
-            if (Model.ProgressivelyEnhance)
-            {
-                <div id="page-actions-button"></div>
-            }
-            else
-            {
-                <div
-                    data-launch-modal
-                    data-modal-name="modal-save-images"
-                    data-element-class-name="@Model.SaveClassName"
-                    data-button-label="Save chart images"
-                    data-modal-title="Save chart images"
-                    data-element-title-attr="@Model.SaveTitleAttr"
-                    data-cost-codes-attr="@Model.CostCodesAttr"
-                    data-show-titles="true"
-                    data-show-progress="true"
-                    data-file-name="@Model.SaveFileName"
-                    data-save-event-id="save-chart-as-image"
-                    data-main-content-id="main-content"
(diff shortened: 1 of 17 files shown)


web/src/Web.App/Views/Shared/Components/PageActions/Default.cshtml
jrabbott, line 1: With these change I'm unclear on the purpose of `SaveButtonVisible` and `DownloadButtonVisible`? Is the intention to remove these in a future PR? or am I missing something?

WolfyUK: There are other consumers of this view component outside the School spending priorities page. This consumer only has the 'Save all' button, but there are different combinations elsewhere. This came out of the Custom Reporting epic. I'm guessing at some point all pages will support both buttons?

jrabbott: Ah yes. That's the bit I'd forgotten.

Suggestion - maybe an improvement here could be, instead of prop for visible on each button we pass in a list/hashset of buttons to display (similar approach to finance tools and resources view components)?

jrabbott (approved):
Looks good - suggested improvement, however not necessary.

web/src/Web.App/Views/Shared/Components/PageActions/Default.cshtml
WolfyUK: Great idea. I'll get this merged and follow up with additional work

PR #2437, opened 2025-06-02
API versioning

jrabbott:
Introduces API versioning. By implementing versioning, the API can support multiple versions concurrently, facilitating smoother transitions and backward compatibility as the API evolves.

Key Changes:

- API Versioning Implementation: Introduces a structured approach to handle multiple API versions, allowing for the coexistence of different API iterations.
- Routing Adjustments: Modifies routing configurations to direct requests to the appropriate versioned function based on the API version specified in the request.
- Function Updates: Refactors existing functions to accommodate versioning, ensuring that each version operates independently without interference.
- Testing Enhancements: Adds comprehensive tests to verify the correct functioning of each API version, ensuring reliability and stability across versions.

These enhancements lay the groundwork for future API developments, enabling the us to introduce new features or changes without disruption.

### DIFF
--- a/platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
+++ b/platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.Functions.Worker.Http;
 using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Attributes;
+using Microsoft.OpenApi.Models;
 using Platform.Api.Establishment.Features.Trusts.Handlers;
 using Platform.Api.Establishment.Features.Trusts.Models;
 using Platform.Functions;
@@ -19,6 +20,7 @@ public class GetTrustFunction(IVersionedHandlerDispatcher<IGetTrustHandler> disp
     [OpenApiSecurityHeader]
     [OpenApiOperation(nameof(GetTrustFunction), Constants.Features.Trusts)]
     [OpenApiParameter("identifier", Type = typeof(string), Required = true)]
+    [OpenApiParameter(Platform.Functions.Constants.ApiVersion, Type = typeof(string), Required = false, In = ParameterLocation.Header)]
     [OpenApiResponseWithBody(HttpStatusCode.OK, ContentType.ApplicationJson, typeof(Trust))]
     [OpenApiResponseWithBody(HttpStatusCode.BadRequest, ContentType.ApplicationJsonProblem, typeof(ProblemDetails))]
     [OpenApiResponseWithoutBody(HttpStatusCode.NotFound)]


jrabbott:
@WolfyUK thoughts?

platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
```diff
@@ -1,36 +1,37 @@
...
+namespace Platform.Api.Establishment.Features.Trusts.Functions;
 
-public class GetTrustFunction(ITrustsService service)
+public class GetTrustFunction(IVersionedHandlerDispatcher<IGetTrustHandler> dispatcher)
 {
     [Function(nameof(GetTrustFunction))]
     [OpenApiSecurityHeader]
     [OpenApiOperation(nameof(GetTrustFunction), Constants.Features.Trusts)]
     [OpenApiParameter("identifier", Type = typeof(string), Required = true)]
     [OpenApiResponseWithBody(HttpStatusCode.OK, ContentType.ApplicationJson, typeof(Trust))]
```
WolfyUK, line 22: It would be handy to also include the optional header in the affected OpenApi specs, e.g.:

```csharp
[OpenApiParameter(Platform.Functions.Constants.ApiVersion, Type = typeof(string), Required = false, In = ParameterLocation.Header)]
```

platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
```diff
@@ -1,36 +1,37 @@
...
         string identifier,
         CancellationToken cancellationToken = default)
     {
-        var response = await service.GetAsync(identifier, cancellationToken);
-        if (response == null)
-        {
-            return req.CreateNotFoundResponse();
-        }
+        var version = req.ReadVersion();
+        var handler = dispatcher.GetHandler(version);
```
WolfyUK, line 31: Consider refactor to something like `GetHandlerOrThrow()` with a custom exception picked up in middleware to return the 'unsupported version' response.

platform/src/apis/Platform.Api.Establishment/Features/Trusts/TrustsFeature.cs
```diff
@@ -13,6 +15,8 @@ public static class TrustsFeature
     public static IServiceCollection AddTrustsFeature(this IServiceCollection serviceCollection)
     {
         serviceCollection
+            .AddSingleton<IGetTrustHandler, GetTrustV1Handler>()
+            .AddSingleton<IVersionedHandlerDispatcher<IGetTrustHandler>, VersionedHandlerDispatcher<IGetTrustHandler>>()
```
WolfyUK, line 19: The answer might be 'no', but is there a way to define the DI to use `IVersionedHandler` with the same generic type for both the interface and concrete class to prevent having to explicitly register in each 'feature'? e.g.:

```csharp
serviceCollection
    .AddSingleton(typeof(IVersionedHandlerDispatcher<>), typeof(VersionedHandlerDispatcher<>))
```

platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
WolfyUK: How about a `TryGetHandler()` instead? 🙃

```csharp
if (dispatcher.TryGetHandler(version, out var handler)) {
    return await handler.HandleAsync(req, identifier, cancellationToken);
}

return await req.CreateUnsupportedVersionResponseAsync(cancellationToken);
```

platform/src/apis/Platform.Api.Establishment/Features/Trusts/TrustsFeature.cs
jrabbott: It needs to be specific - each function end point needs to have it's own interface that implements `IVersionedHandler` to handle the routing parameter specific to that endpoint.

The dispatcher needs to return that specific type for that end point, to pass those parameters too,

platform/src/apis/Platform.Api.Establishment/Features/Trusts/Functions/GetTrustFunction.cs
jrabbott: Nice suggestion - however you'd then need an additional null check on the out parameter.

Having compared the two, the current implementation feels the simplest - happy to jump on a call and discuss?

WolfyUK (approved):
🚀 Good one! Not sure if others have feedback too.

jrabbott:
Thanks @WolfyUK. @J05h-L / @Faizan-ahmad00 waiting on your feedback before merging.

jrabbott:
@Faizan-ahmad00 a few more questions.... it's great that we're thinking ahead about test strategy across versions.

How do we see the trade-off between duplication and reuse when it comes to testing versioned APIs? For example, if two versions share behaviour now but might diverge later, should we lean toward shared tests to avoid duplication, or separate tests to protect version integrity?

Also, what risks do we see if we tightly couple test logic across versions? Could that introduce unintentional regressions in older versions?

Faizan-ahmad00:
> @Faizan-ahmad00 a few more questions.... it's great that we're thinking ahead about test strategy across versions.
> 
> How do we see the trade-off between duplication and reuse when it comes to testing versioned APIs? For example, if two versions share behaviour now but might diverge later, should we lean toward shared tests to avoid duplication, or separate tests to protect version integrity?
> 
> Also, what risks do we see if we tightly couple test logic across versions? Could that introduce unintentional regressions in older versions?

@jrabbott Thanks totally agree, these are really good points. I was thinking along similar lines especially about the balance between duplication vs. islolating versions. 
I think these are best explored in more depth outside the PR, so I will take some time to think through it properly but for now i am inclined towards isolation. Would be good to align on how we want to scale the test strategy as versions start to diverge.

Appreciate you bringing this up. Definitely worth a proper discussion.

PR #2493, opened 2025-06-19
feat: `LaNationalRankView` to show charts for both outturn as a percentage of funding and of budget 

J05h-L:
### Context
[AB#265025](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/265025) [AB#265455](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/265455) [AB#265456](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/265456)

- adds gds tag component and left align chart mode toggle within each panel.
- separate partial for funding and budget
- chart amended to be used by both

### Guidance to review 

Can be validated using `npm run dev` in front-end components establishment platform api pointing to d12

Another pass is required for content tweaks. This will be addressed in a future PR.

My intention is to prevent users from encountering an empty tab if data retrieval fails. To handle this scenario, the `DataWarning` component has been used. Open to alternative suggestions if there’s a more effective approach.

- [x] Your branch has been rebased onto feature branch


WolfyUK:
This looks great, @J05h-L ! Just a couple of comments:

1. What is the difference between the two new 'section's? It may be more manageable to use props to define underlying data sources, labels, etc. if they are very similar.
1. It looks like both API calls take place as soon as the page is loaded. Is there any way to delay until the tab is opened, as done elsewhere?

J05h-L:
@WolfyUK Once the changes are made for the `NationalRankApi` The two sections will have different api calls and therefore different data and I was worried that might be a bit too busy all staying in the one place. I thought it made sense to keep them separate. Happy to change that but that was the motivation.

I think @jrabbott has just replied regarding point 2

J05h-L:
@WolfyUK 

> Will the data be different though, requiring changes to the dataPoint mapping?

no I expect that to be the same. I will mark back as draft and wire this up while moving everything back to one place

J05h-L:
@WolfyUK You're right this is tidier, refactored. This covers the wiring up now as well so AB#265456 can be closed after a bump to front end. Just testing now. Will mark as ready for review shortly

WolfyUK (approved):

PR #2668, opened 2025-07-28
Feature/296893/cfr-input-scheme-additions

murdo-hippo:
## 🧾 Summary
To test CFR IT Spend changes for 2025 data, all ancillary data schemas for 2025 must be added.
Note that as of this PR, all of the schemas are the same as 2024 and all 2025 ancillary data files are the same as 2024


murdo-hippo:
For most schema, there's a check for the current year with a "default" schema fallback, it doesn't take the "latest". Notably if the CFR and AAR years are different, then [2 versions of the ancillary data get brought in.](https://github.com/DFE-Digital/education-benchmarking-and-insights/blob/49061e0e32b01153429934bb7e165084da26d34c/data-pipeline/src/pipeline/main.py#L826) This ancillary data includes AAR data so schemas for everything needs to be defined.

I think we could probably handle the getting of Schemas differently to how we do now and it'd be clearer. More refactoring for the pile.

data-pipeline/src/pipeline/input_schemas/aar_academies.py
```diff
@@ -197,6 +197,73 @@
         "BAI101-T": "float",
         "BAI011-A": "float",
     },
+    2025: {
```
J05h-L, line 200: I'm a little late to this one sorry. I'd expect we could have skipped creating a 2025 schema for `aar_academies.py` and `aa_central_services.py`. A run using something like the message below should be fine and use the existing 2024 data.

`{
  "type": "default",
  "runId": 2025,
    "year": {
        "aar": 2024,
        "cfr": 2025,
        "bfr": 2025,
	"s251": 2024
    }
}`

Possible I've missed something but defining a new schema for aar 2025 doesn't seem right.

jrabbott: It doesn't matter too much at this point, as you've said.

PR #2700, opened 2025-08-01
2025 CFR file tweaks

murdo-hippo:
## 🧾 Summary
* The new IT spend breakdown columns needed two spaces between the code and the text description
* As I was debugging the federation funnies breaking the pipeline, I clearcoded the variable names in `join_federations`. It's an improvement so I have left it in


murdo-hippo:
branch name wrong

PR #2701, opened 2025-08-01
Feature/269159/verify 2025 cfr file

murdo-hippo:
## 🧾 Summary

- The new IT spend breakdown columns needed two spaces between the code and the text description
- As I was debugging the federation funnies breaking the pipeline, I clearcoded the variable names in join_federations. It's an improvement so I have left it in


murdo-hippo:
The federations issue results in duplicates in the eventual maintained schools table. I have a line of code to dedupe this maintained schools table, which allows me to run the pipeline and unblocks me checking all the ancillary files. In parallel Ben is changing the CFR file so the duplicates are dealt with.

PR #2732, opened 2025-08-15
Remove outdated e2e data pipeline tests

murdo-hippo:
## 🧾 Summary
* Tests are from Nov 2024 and fail when triggered in CICD
* As far as I know the implementation never got done


jrabbott (approved):
Would be good to add some in!

PR #2734, opened 2025-08-15
Test CICD environment works for pdf generation enhancements

murdo-hippo:
## 🧾 Summary
* set more config so tables render nicely in PDF
* add libraries so the config works

### DIFF
--- a/pipelines/documentation/build.yaml
+++ b/pipelines/documentation/build.yaml
@@ -6,11 +6,10 @@ trigger:
   branches:
     include:
       - main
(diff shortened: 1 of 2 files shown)


pipelines/documentation/build.yaml
```diff
@@ -6,6 +6,7 @@ trigger:
   branches:
     include:
       - main
+      - feature/improve-doc-pdf-outputs
```
jrabbott, line 9: Not required, You can manually run from any branch.

jrabbott (approved):

PR #2742, opened 2025-08-19
Vectorise and parallelise RAG calculations

murdo-hippo:
## 🧾 Summary
[AB#266551](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/266551)
* modularised the code even more
* Clearcode rag functions
* User defined and default pathways have been standardised as much as possible
* Bring constants to the top of the file. This makes life easier for when schemas are eventually refactored.
* Vectorise calculation of `find_pupil_close_comparators` using `numpy` instead of `math`. I had to reimplement the algorithm for `math` as it's subtly different to `numpy`'s calculation. This gave about a 25% speedup from 8.2 sec/1k RAG ratings to 6.1 sec/1k RAG ratings
* Add a class `CategoryColumnCache` to cache the pandas masks required to look up comparator sets
* Parallelise RAG calculations on as many cores as the compute has. On my Macbook pro this is 12 cores and all RAG ratings for a set of schools (12k isk) computes in 12 seconds. In practice setting up and dealing with the workers for the parallelisation means that the overall time saving is less, but it still saves about a minute overall. On azure remote compute this will be less as it won't have 12 cores.
* Overall time savings are something like 200sec -> 60 sec
* Redo the tests for these new functions

## 🧪 Testing notes
* I tested the output for the RAG ratings of 2021-2025 before and after these changes using the script I have to hash all the database parquet files. They were the same.
* I also tested triggering a custom comparator set run, and it completed ok

---
from the comparator set refactor:
## 🧾 Summary
[AB#266550](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/266550)
* Vectorise and clearcode comparator sets code
* Create config file to prep for input schemas refactoring
* Execution time on mac goes from ~22 sec to ~18 sec (not consistent)
* New tests for calculation logic

## Notes
* We complete preprocessing, then `prepare_data` in this pipeline does more preprocessing, then persists the repreprocessed data again to parquet. This feels inconsistent but I wanted to make no output changes in this PR so I didn't fix it.
* Theres another potential bug behaviour which I've kept the same when filling comparator sets after distance calculations to avoid making logic changes. There are comments explaining the potential bug, it could result in duplicates in the comparator set.

## 🧪 Testing notes
* Hashes of the comparator set parquet files in blob storage are the same before and after these changes from 2021-2025.


jrabbott:
Could we keep this as draft until CFR is done? Just in case any pipeline changes are needed.

PR #2771, opened 2025-08-27
Bugfix: add non teachers to senior leadership totals

murdo-hippo:
## 🧾 Summary
* Senior leadership figures should involve non teachers, so aggregate the teachers and non teachers.

## 🔍 Reviewer notes
* I have backdated this fix to 2024. That means that non CFR numbers will change.
* Pre-2024 non-teacher figures were not included in workforce census data, so for pre-2024 we won't see any change in numbers.

## 🧪 Testing notes
- I have checked the calculation works for a few rows.
- Pipeline preprocessing passes for 2025, 2024, and 2022


jrabbott:
@murdo-hippo A few points to flag:

- Putting aside the requirement that the aggregation spans all years, the current logic would cause the pipeline to fail if we attempt to run pre-2024 data.
- The pipeline needs to remain backwards compatible so that prior years can always be executed successfully.
- While not all new fixes/features will apply to prior years, historical runs should still work without breakage.
- Previous years should be idempotent (unless there’s an intentional, impacting code change).

PR #2933, opened 2025-09-10
doc: Remove references to other costs direct revenue financing

murdo-hippo:
## 🧾 Summary
[AB#278482](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/278482)
* Update models
* Update cost categories
* Remove double reference to direct revenue financing in income financing


murdo-hippo:
The models documentation markdown page which shows schemas, aggregated columns, and how names change through the pipeline makes me want to sort out the input schemas. Ideally I would like this all in code and readable there rather than a markdown doc which can drift from the code.

I think this is a good over-Christmas task as it'll be a bit of a job and is just refactoring.

jrabbott (approved):

PR #3078, opened 2025-10-07
feat: Persist KS4 banding to db

murdo-hippo:
## 🧾 Summary
[AB#283987](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/283987)
* We already read both KS4 progress indicators and banding (eg Average, Below Average) from the data source, however the banding isn't persisted to the db
* Persists it to the db so we can do the progress indicators work.


jrabbott:
I believe this is a misunderstanding of the requirements/data.

The banding is used to determine the confidence which isn't something we are looking at surface.

PR #4048, opened 2026-04-17
feat: add data tables and toggle between chart/table views to ehcp page

J05h-L:
## 🧾 Summary

[AB#299498](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/299498) [AB#308381](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/308381)

#### Functionality:

- Adds data tables to the new EHCP page.
- Introduces a chart/table toggle (charts currently placeholder text until SSR bar charts are implemented).
- No filtering yet, all categories are shown by default.

#### Implementation:

**Domain**

`EducationHealthCarePlans`

A new domain model representing the API response.
This duplicates `LocalAuthorityNumberOfPlans`, but the existing types aren’t ideal and there is already duplication across LA domain models (`Web.App.Domain.LocalAuthorities` vs `Web.App.Domain.LocalAuthority`).

Proposal: address the duplication across Local Authority in a separate ticket. I’ll raise one after review if agreed.

`EducationHealthCarePlansCategories`

Mirrors the IT Spend pattern.

- Represents category filters
-  Provides `GetTableHeading()` to supply the correct column heading for each category.

**Controllers**

LocalAuthorityEducationHealthCarePlansController now:

- Calls `LocalAuthorityApi.QueryEhcpAsync` to fetch comparator‑set plans.
- Redirects with a `viewAs` query parameter to control chart/table mode.

**View Models**

- `LocalAuthorityEducationHealthCarePlansViewModel` now includes SubCategories and ViewAs.
- `EducationHealthCarePlansComparisonSubCategoriesViewModel` responsible for transforming the raw `EducationHealthCarePlans[]` returned from the API into a set of sub‑category comparison groups suitable for rendering as tables (and later charts). Mirrors the pattern used in IT Spend
- `EducationHealthCarePlansComparisonDatum` created to represent a single row of comparison data (usable for both tables and charts) will be moved to `Domain/Charts` once work to implement the charts is underway.
- `LocalAuthorityEducationHealthCarePlansTableViewModel` added to represent each sub‑category table.
- `BenchmarkingViewModelCostSubCategory<T>` updated with a `TableHeading` property.

**Views**

- New partial for rendering each EHCP table, with a fallback warning if data is null.
- Added a form to toggle between chart and table view.
- Conditional rendering: tables for each selected sub‑category or placeholder chart text.

## ✅ Checklist

## 🔍 Notes

Table headings have been agreed with Design; they were not all originally specified in the Figma. Some alignment adjustments will be needed for the column headers, but it makes more sense to revisit this once the page is closer to its final state.

This is a slightly larger PR than ideal sorry but a chunk of these lines are within the integration tests. Happy to split if the reviewer feels that would be helpful

### DIFF
--- a/web/src/Web.App/Controllers/LocalAuthorityEducationHealthCarePlansController.cs
+++ b/web/src/Web.App/Controllers/LocalAuthorityEducationHealthCarePlansController.cs
@@ -1,11 +1,13 @@
 ﻿using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
 using Web.App.Attributes;
-using Web.App.Domain;
+using Web.App.Domain.Charts;
+using Web.App.Domain.LocalAuthorities;
 using Web.App.Infrastructure.Apis;
 using Web.App.Infrastructure.Extensions;
 using Web.App.Services;
 using Web.App.ViewModels;
+using LocalAuthority = Web.App.Domain.LocalAuthority;
 
 namespace Web.App.Controllers;
 
@@ -19,7 +21,9 @@ public class LocalAuthorityEducationHealthCarePlansController(
     : Controller
 {
     [HttpGet]
-    public async Task<IActionResult> Index(string code)
+    public async Task<IActionResult> Index(
+        string code,
+        Views.ViewAsOptions viewAs = Views.ViewAsOptions.Chart)
     {
         using (logger.BeginScope(new { code }))
         {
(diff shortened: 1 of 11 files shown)


mattb-hippo (approved):
We will have to think carefully about the communication/integration between the form declared in `web/src/Web.App/Views/LocalAuthorityEducationHealthCarePlans/Index.cshtml` and the form we declare in the partial which will hold the left-hand filters.

But other than that, this is a great basis to build the rest of the page from.

PR #4054, opened 2026-04-21
feat: stitch EHCP charts into new page

mattb-hippo:
Fixes [AB#308379](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/308379) (Task within [AB#299498](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/299498))

Screenshot 2026-04-21 at 14 59 02

## ✅ Checklist

I've created a new view model for the charts partial (`LocalAuthorityEducationHealthCarePlansChartViewModel`)
which is actually identical to the existing table partial (`LocalAuthorityEducationHealthCarePlansTableViewModel`)
and I wasn't 100% sure that was the correct approach. Perhaps we could instead use the same class for both
purposes?

ALso, I've deliberately deferred any test writing for this.  I think any tests written now will be so badly affected once filtering is introduced that spending time on it now will be a bit of a waste.

### DIFF
--- a/web/src/Web.App/Controllers/LocalAuthorityEducationHealthCarePlansController.cs
+++ b/web/src/Web.App/Controllers/LocalAuthorityEducationHealthCarePlansController.cs
@@ -17,6 +17,7 @@ namespace Web.App.Controllers;
 public class LocalAuthorityEducationHealthCarePlansController(
     ILogger<LocalAuthorityEducationHealthCarePlansController> logger,
     ILocalAuthorityApi api,
+    IChartRenderingApi chartRenderingApi,
     ILocalAuthorityComparatorSetService comparatorSetService)
     : Controller
 {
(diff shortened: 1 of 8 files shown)


J05h-L (approved):
Looks good. I particularly like `EducationHealthCarePlanProperties` to avoid the magic string (that I introduced previously).

Agreed that adding tests now will just cause rework.

I also can't see a reason to keep `LocalAuthorityEducationHealthCarePlansChartViewModel` and `LocalAuthorityEducationHealthCarePlansTableViewModel` separate. 

Not worth blocking this imo but probably worth us refactoring to just use one view model for the sub category as part of the work on this story. Could be added as a TODO on AB#309435?

mattb-hippo:
Will add some commentary to AB#309435

PR #4056, opened 2026-04-22
feat: add `SENSupport` field to `LocalAuthorityFinancial`

murdo-hippo:
## 🧾 Summary
To support a new filter on the high needs page for local authorities which normalises spend using SEN support figures, per-LA SEN support values are needed.

Fixes [AB#309836](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/309836)

## ✅ Checklist


jrabbott:
@murdo-hippo please could you use the PR template. If the template isn't working, then lets improve it.

murdo-hippo:
Done ✅

core-infrastructure/src/db/Core.Database/Scripts/135-AlterLocalAuthorityFinancialAddSENSupport.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE dbo.LocalAuthorityFinancial
```
J05h-L: I think this should be wrapped in an `IF EXISTS` guard. I appreciate that this isn’t applied consistently across our scripts but I think it should be, it keeps deployments safer in environments where the table may not exist yet.

Happy to hear others’ opinions on this, as I think it’s worth standardising the approach across the codebase.

J05h-L (commented):
Looks good. I really like the docstring, it gives helpful context on why `_aggregate_sen_support_to_la_level()` behaves differently to `_aggregate_fbit_pupil_numbers_to_la_level()`, which will make life easier for future maintainers.

I’ve added a separate comment on the database script. In my view it should use an `IF EXISTS` guard, and I think it’s worth us standardising that pattern across the codebase so I'm keen to hear opinions on that.

core-infrastructure/src/db/Core.Database/Scripts/135-AlterLocalAuthorityFinancialAddSENSupport.sql
murdo-hippo: Is the argument for the `IF EXISTS` guard that migration scripts might be run more than once and fail a deployment? I think I'm coming from the POV that each script should be run ONLY once and migration script errors are desired as a signal that we need to fix something (plus it makes scripts simpler and easier to parse).

Standardisation sounds good, we could have a tech debt item for that?

J05h-L: You’re right that these scripts should only run once. Personally I still lean toward the more defensive approach, but I do hear the argument for keeping things simple. I think my main concern is the inconsistency across our scripts rather than this specific case.

Happy to approve if we can get something into the backlog to discuss and agree a standard for future changes.

jrabbott: From my perspective, we had a standard pattern and have drifted away from it. We shouldn't be added debt if we aren't following our standards.

murdo-hippo: What's the standard pattern sorry? I'll change this to align with it

jrabbott: See the earlier scripts.

jrabbott: I think we should be moving back towards migrations being idempotent and defensive by default, even if we intend them to run only once. A few reasons why:

**Deployments aren’t always clean or linear**
Even with the journey table ensuring scripts only run once, failures mid-deployment can leave things in a partially applied state. A defensive script gives us a safer re-run path without needing manual intervention.

**Partial execution is a real risk**
If a script fails halfway through (e.g. table created but index not, or column added but data not migrated), a non-idempotent script can make recovery harder. Guards like IF EXISTS / IF NOT EXISTS help us resume safely.

**Environments drift**
Lower environments (and occasionally prod via hotfixes or manual changes) don’t always perfectly match expectations. Defensive scripts reduce the chance of unexpected failures due to that drift.

**Operational safety > purity**
I agree in principle that scripts should run once and fail loudly if something’s wrong, and that’s totally fine (even helpful) in dev. But in production, the cost of a failed deployment is high. These scripts need to be production-ready, and a bit of defensiveness buys us resilience.

**Consistency matters**
Regardless of which direction we take, the bigger issue is having a clear, enforced standard. Right now we’re inconsistent, which is where the real risk and confusion comes in.


_...sorry hit return too quickly on my previous message_

J05h-L: I was thinking of something like this

```
IF EXISTS(SELECT 1
          FROM INFORMATION_SCHEMA.TABLES
          WHERE table_name = 'LocalAuthorityFinancial')
    BEGIN
    ALTER TABLE dbo.LocalAuthorityFinancial
        ADD SENSupport DECIMAL(16,2) NULL
    END;
 ````
 
 which is the approach we have in `130-AlterLAFinancialAndLANonFinancialToAddPupilNumbers.sql` and `126-AlterNonFinancialTable.sql`

J05h-L: Although sorry I missed this while first looking at the PR. Should this be `LocalAuthorityFinancial` or `LocalAuthorityNonFinancial`. SEN support feels like it should be part of the non financial data?

murdo-hippo: You're right Josh! a couple of changes incoming

J05h-L (approved):

PR #4068, opened 2026-04-23
feat: add save chart images and download page data to ehcp page

J05h-L:
## 🧾 Summary

[AB#308380](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/308380) [AB#308380](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/308380)

### ✨ Feature Details
**Functionality:**

This PR updates the ehcp page by adding "page‑action" buttons for saving chart images and downloading the underlying data. It also introduces a new controller action that generates a CSV and bundles it into a ZIP file for download.

 **Implementation:**

Added a new .page-actions-wrapper-horizontal class to align page‑action buttons horizontally on larger screens and stack them vertically on smaller breakpoints.

Updated the view to render the Download page data button within the new layout wrapper and only render the Save chart images button when the page is in Chart view.

Integrated the existing PageActions enhancement component, including the required partial in the scripts section, to expose the correct element IDs, filenames, etc for chart export.

image

## ✅ Checklist

## 🔍 Notes

Possible further work will be required to ensure the styling correctly matches the designs but this PR is intended to cover all functionality and the basic styling for these buttons

### DIFF
--- a/web/src/Web.App/Views/LocalAuthorityEducationHealthCarePlans/Index.cshtml
+++ b/web/src/Web.App/Views/LocalAuthorityEducationHealthCarePlans/Index.cshtml
@@ -1,6 +1,7 @@
 @using Web.App.Domain.Charts
 @using Web.App.Extensions
 @using Web.App.ViewModels
+@using Web.App.ViewModels.Enhancements
 @model Web.App.ViewModels.LocalAuthorityEducationHealthCarePlansViewModel
 @{
     ViewData[ViewDataKeys.Title] = PageTitles.LocalAuthorityEducationHealthCarePlans;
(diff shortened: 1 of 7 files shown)


web/src/Web.App/Views/LocalAuthorityEducationHealthCarePlans/Index.cshtml
```diff
@@ -101,4 +119,18 @@
             </section>
         }
     </div>
-</div>
\ No newline at end of file
+</div>
+
+@section scripts
```
mattb-hippo, line 124: How does this degrade if the user doesn't have javascript enabled?  Do we also disable the button?

mattb-hippo (approved):
One comment for my curiosity, but otherwise all good.

web/src/Web.App/Views/LocalAuthorityEducationHealthCarePlans/Index.cshtml
J05h-L: The div with id `page-actions-button` will not contain a button, it will just be an empty div in the DOM. Will look like this for the user

image

_____

I suppose we could add a fallback in a noscript with a disabled button?

PR #4080, opened 2026-04-28
feat: Bootstrap pipeline setup into Docker compose

murdo-hippo:
## 🧾 Summary
The setup of the data pipeline locally has lots of steps. It's in our interest to get this as easy as possible, for ease of onboarding and testing going forward.

This simplifies the setup of the data pipeline in the project level `docker-compose` by bootstrapping setup for Azureite (creating containers and queues) and MSSQL (running db migrations) into init containers.

## ✅ Checklist
- [x] Relevant documentation updated _(if applicable)_.


jrabbott (commented):
Nice approach using the side-cars. A few comments/recommendations

1. Robustness: Resolve Race Conditions
The azurite-init and mssql-init services currently use condition: service_started. Since both Azurite and SQL Server take several seconds to fully boot and bind to their ports, the initialisation scripts are likely to fail on the first attempt if they run before the engine is ready.
    * Recommendation: Add a healthcheck to the azurite and mssql services.
    * Recommendation: Update the depends_on in the init services to use condition: service_healthy. This ensures a "one-click" startup that doesn't require manual restarts.

2. Stability: Prevent Host Volume Corruption
The db-migrate service mounts the entire repository root into a .NET SDK container to execute dotnet run. 
    * The Issue: The Linux container will write Linux-compiled assets into the host's bin and obj folders. On a Windows or macOS host, this leads to file lock issues, platform-specific compilation errors, and "stale" binary conflicts in the IDE.
    * Recommendation: Instead of mounting the repository, create a dedicated multi-stage Dockerfile for the Core.Database project. This keeps the build process isolated inside the container and prevents polluting the host filesystem.

3. Strategy: Validation of the Sidecar Pattern
The use of a separate mssql-init service is the correct architectural choice here.
    * Observation: While we could create a custom SQL image with a wrapper script, the "sidecar" approach is cleaner for MSSQL/Azure SQL Edge. It allows us to use the official image directly and provides a clear service_completed_successfully event that the db-migrate service can depend on. It’s a more maintainable and "Docker-native" orchestration pattern.

jrabbott (commented):
[suggestion] Centralise credentials using a .env file. 

To keep the configuration DRY, we should move the SQL credentials and database names into environment variables.

Currently, the password  and the database name data are hardcoded across four different services (mssql, mssql-init, db-migrate, and the azurite-init connection string).

murdo-hippo:
That's done! The healthchecks were a bit fiddly as the containers have limited tooling in them (and I wanted to avoid having to add any more). I've just gone for a fairly minimal approach of checking the ports which should be active and listening for the application are open.

core-infrastructure/src/db/Core.Database/Dockerfile.migrate
```diff
@@ -0,0 +1,15 @@
+# build/publish stage
+FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
+WORKDIR /src
+
+COPY . .
```
jrabbott: Feels overkill copying the whole repo? Just the core-infrastructure folder is needed.

docker/.env.example
jrabbott, line 1: Could we add the vars for redis container too. Thanks

murdo-hippo:
Ok thanks I've made those tweaks, that makes it a bit neater.

jrabbott (dismissed):
This works a dream. One minor thing, take or leave.

 Could we add `CONTAINERS` and `QUEUES` to the environment section of the `azurite-init`, then modify the shell command to iterate over the new environment variables ($$CONTAINERS and $$QUEUES) instead of the hardcoded strings?

murdo-hippo:
I can do that, it's quick. Other ideas I had:
* Include the data loading to azurite as part of the docker startup (there'd be a step in the readme to put the `default` folder into the repo to feed it
* Include config to run the pipeline in debug mode, so that a debugger can be mounted on to the running container to identify issues

jrabbott (approved):

PR #4148, opened 2026-05-08
feat: add data sources to high needs spending benchmarking

mattb-hippo:
Fixes [AB#295179](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/295179) [AB#311590](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/311590)

Screenshot 2026-05-08 at 15 36 16

## ✅ Checklist

Because the S251 year needs to be displayed in the data source block I made a specific decision here to add the `IFinanceService` into the high needs controller via dependency injection to obtain the year info.  That service does already get used in the page, in the `DataSourceViewComponent` which does mean that this page is calling `IFinanceService.GetYears()` twice. The alternative was to refactor that view component so that the year information is passed in, rather than have it fetch it itself, but that would have entailed modifying all pages that use the component.

### DIFF
--- a/web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
+++ b/web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
@@ -138,6 +138,8 @@
                         {
                             @await Html.PartialAsync("_HighNeedsSpendingTable", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!))
                         }
+                        
+                        @await Html.PartialAsync("_HighNeedsSpendingDataSources", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!, Model.Year))
                     </section>
                 }
             </section>
--- a/web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
+++ b/web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
@@ -132,7 +132,7 @@
 
                         @if (Model.ViewAs == Views.ViewAsOptions.Chart)
                         {
-                            @await Html.PartialAsync("_HighNeedsSpendingChart", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!))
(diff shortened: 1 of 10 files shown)


web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
```diff
@@ -132,12 +132,14 @@
...
                         {
-                            @await Html.PartialAsync("_HighNeedsSpendingChart", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!))
+                            @await Html.PartialAsync("_HighNeedsSpendingChart", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!, Model.Year))
                         }
                         else
                         {
                             @await Html.PartialAsync("_HighNeedsSpendingTable", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!))
                         }
+                        
+                        @await Html.PartialAsync("_HighNeedsSpendingDataSources", new LocalAuthorityHighNeedsSpendingDataViewModel(subCategory, Model.ResultAs, Model.Type, Model.Code!, Model.Year))
```
J05h-L: The data‑source elements need to live inside the chart/table partials. It’s less important for tables, but for charts these sources must sit within the costs-chart-container so the save‑image enhancement captures them correctly.

J05h-L (changes requested):
Looks great, just one small comment to make sure the save image enhancement also picks up the data‑source elements. Other than that, the implementation is solid and I agree with the decisions outlined in the description.

web/src/Web.App/Views/LocalAuthorityHighNeedsSpending/Index.cshtml
mattb-hippo: I will get on with that change now, good catch, thank you.

mattb-hippo: Updated in latest commit

J05h-L: Nice one

J05h-L (approved):

PR #4149, opened 2026-05-11
test: add integration tests for high needs spending

J05h-L:
## 🧾 Summary

[AB#295179](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/295179) [AB#311303](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/311303)

### ✨ Feature Details
**Functionality:** 

- no changes to functionality. 
  - to assist with selection of elements in tests a `data-testid` has been added for each sub category 
  - an error in the `Url.Action` for download data has been fixed. ie now `(int)Model.ResultAs`, instead of `Model.ResultAs,`

**Implementation:**

- add integration test cover for this page. 
  - table contents are asserted for all values
  - subcategories and submission type for filter options
  - table/chart view as options and also resultAs dimensions are also covered.

## ✅ Checklist

### DIFF
--- a/web/tests/Web.Integration.Tests/Pages/LocalAuthorities/WhenViewingHighNeedsSpending.cs
+++ b/web/tests/Web.Integration.Tests/Pages/LocalAuthorities/WhenViewingHighNeedsSpending.cs
@@ -9,16 +9,16 @@ using Xunit;
 
 namespace Web.Integration.Tests.Pages.LocalAuthorities;
 
-public class WhenViewingHighNeedsSpendingPlans(SchoolBenchmarkingWebAppClient client) : PageBase<SchoolBenchmarkingWebAppClient>(client)
+public class WhenViewingHighNeedsSpending(SchoolBenchmarkingWebAppClient client) : PageBase<SchoolBenchmarkingWebAppClient>(client)
 {
     #region Tests
 
     [Fact]
     public async Task CanDisplay()
     {
-        var (page, authority, plans) = await SetupNavigateInitPage();
+        var (page, authority, expenditures) = await SetupNavigateInitPage();
 
-        AssertPageLayout(page, authority, plans);
+        AssertPageLayout(page, authority, expenditures);
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


web/tests/Web.Integration.Tests/Pages/LocalAuthorities/WhenViewingHighNeedsSpending.cs
```diff
@@ -0,0 +1,462 @@
...
+using AngleSharp.Html.Dom;
+using AutoFixture;
+using Web.App.Domain.Charts;
+using Web.App.Domain.LocalAuthorities;
+using Web.App.Extensions;
+using Xunit;
+
+namespace Web.Integration.Tests.Pages.LocalAuthorities;
+
+public class WhenViewingHighNeedsSpending(SchoolBenchmarkingWebAppClient client) : PageBase<SchoolBenchmarkingWebAppClient>(client)
```
mattb-hippo, line 12: Should we be using `SchoolBenchmarkingWebAppClient` in a Local Authority test

mattb-hippo (changes requested):

web/tests/Web.Integration.Tests/Pages/LocalAuthorities/WhenViewingHighNeedsSpending.cs
J05h-L: Good spot. We’ve ended up using `SchoolBenchmarkingWebAppClient` across the tests simply because it’s the only concrete client we currently have. It became the “default” even in tests that aren’t school specific. I used it here without really thinking about it for that reason.

Maybe we should introduce something like a `LocalAuthorityBenchmarkingWebAppClient` so the tests reflect the correct auth context? It would be worth looking at this more broadly to tidy things up across the suite.

mattb-hippo: Agreed, it's definitely worth addressing, but not something to stall this PR over. I'll raise a tech debt ticket.

mattb-hippo (approved):
Based on PR comments, happy to approve with tech debt raised

PR #4153, opened 2026-05-11
feat: Guard local development of the data pipeline from secret leakage through LLMs

murdo-hippo:
## 🧾 Summary
Move secret variables out of the repository filesystem to prevent secret leakage though use LLM coding tools. The assumption here is that LLM coding tools are generally restricted to the repository they are working on.

## ✅ Checklist
- [x] Relevant documentation updated _(if applicable)_.


J05h-L:
Looks good, docs cover the setup well. Just one observation, it might be worth including the notes from the PR description in the set up instructions? Seems like a worthwhile thing to call out. What do you think?

Otherwise this branch has gone a bit stale so looks like it needs conflicts resolving but otherwise happy to approve.

PR #4160, opened 2026-05-13
chore: bump front-end dependencies

J05h-L:
## 🧾 Summary

Updates the npm dependencies for the front-end-components project

[AB#312542](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/312542)
[AB#312551](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/312551)

### ⛓️ Tech Debt & Refactor Details
**Major Changes:**

Brings multiple front-end-components dependencies up to their latest compatible versions.

**Benefits:**

Addresses several security vulnerabilities and ensures the project remains stable and up to date.

## ✅ Checklist
- [x] Code follows project coding standards (e.g., Trunk-based guidelines, small PR size).
- [ ] Unit and integration tests added/updated _(if applicable)_.
- [x] Tested by running locally _(or docs render correctly locally)_.
- [ ] Relevant documentation updated _(if applicable)_.
- [ ] Screenshots or Demo included _(for UI/frontend changes)_.

## 🔍 Notes

This PR should close the following dependabot PRs
- #4027
- #4025
- #4024
- #4023
- #4021
- #4019
- #4002
- #4000
- #3998
- #3919

PR #4167, opened 2026-05-14
fix: pin vite to ^7.3.3 and @vitejs/plugin-react to ^5.2.0 to fix errors with AutoComplete component

J05h-L:
## 🧾 Summary

Updates front-end-components to use the latest `Vite` `7.x` `^7.3.3` and `@vitejs/plugin-react` to use `^5.2.0` releases. This avoids breaking changes introduced in Vite 8.x that caused runtime failures in the `AutoComplete` component.

[AB#312542](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/312542)
[AB#312609](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/312609)

### 🛠️ Bug Fix Description
**Root Cause:**
Following #4160 and the subsequent bump to the `front-end` package in Web.App on #4164 runtime errors were observed on the `comparators/create/by/name` pages. These pages use the `SchoolInput` and `TrustInput` component both of which depend on the `AutoComplete` component.

Upgrading to Vite 8.x introduced changes in the React transform pipeline that resulted in invalid JSX element shapes at runtime. This manifested as warnings such as:
`Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object.` when rendering the `AutoComplete` component.

**Changes:**

Pinned Vite to the latest 7.x release
Pinned @vitejs/plugin-react to the 5.2.0
Replaces the earlier Vite 8.x upgrade introduced #4160
Ensures we avoid the security vulnerabilities present in the older Vite version we were previously using, while also resolving the runtime errors.
This stabilises the build and restores correct behaviour for all components relying on `AutoComplete`.

## ✅ Checklist

## 🔍 Notes

There are a few potential long term approaches we could take following this fix:

- **Investigate AutoComplete for Vite 8 compatibility**. Investigate the AutoComplete implementation to understand why it breaks under Vite 8.x. Other components appear unaffected, but nothing obvious stands out from the Vite 8 changelog. This may require deeper inspection of how the component constructs JSX or interacts with the React transform pipeline.
- **Remain on Vite 7.x and ignore future major updates**. We could choose to pin to the latest secure 7.x release and silence Dependabot major bumps. This keeps the build stable with minimal engineering effort.
- **Move towards using the Vue components in Web.App** This aligns with the direction taken for the initial search experience. I haven't fully considered the work involved here or the feasibility but it’s worth noting as a option.

My recommendation is to merge this fix to stabilise the current behaviour, then raise a follow‑up ticket to evaluate these options and decide on the longer‑term direction. Open to alternative suggestions if reviewers see a better path.

### DIFF
--- a/front-end-components/index.html
+++ b/front-end-components/index.html
@@ -80,7 +80,7 @@
       data-highlight="143310"
       data-sort-direction="asc"
     ></div>-->
-  <!--<form method="get">
+  <form method="get">
       <div id="school-suggester" data-exclude="123456,123457"></div>
       <div
         id="la-suggester"
@@ -95,7 +95,7 @@
         data-exclude="06511936,07375627"
       ></div>
       <button type="submit">Submit</button>
-    </form>-->
+    </form>
   <!--<div id="compare-your-trust" 
       data-id="10279605"
       data-show-high-exec="true"
@@ -171,7 +171,7 @@
       <circle r="45" cx="50" cy="50" fill="purple" />
     </svg> -->
   <!--<div id="historic-data-high-needs" data-code="201"></div>-->
-  <div id="benchmark-data-high-needs" data-code="206" data-set='["202", "203", "204", "205"]' data-years-label="Section 251 (test)" data-glossary-url="#"></div>
+  <!-- <div id="benchmark-data-high-needs" data-code="206" data-set='["202", "203", "204", "205"]' data-years-label="Section 251 (test)" data-glossary-url="#"></div> -->
   <script type="module" src="/src/main.tsx"></script>
 </body>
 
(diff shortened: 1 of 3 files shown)


mattb-hippo (approved):
Agree with the suggestion to raise a follow‑up ticket to evaluate options for next steps and decide on the longer‑term direction

J05h-L:
AB#312826 raised to address the compatibility with Vite 8.x and the AutoComplete component with a long term solution

PR #4177, opened 2026-05-18
chore: refactor comparator selection controller for legibility

mattb-hippo:
Fixes [AB#300765](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/300765) [AB#313046](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/313046)

- rewritten the main switch statement to be a little more expansive, include a bit more whitespace, and more legible
- reordered the return statements at the end to be a little more intuitive
- change the comparator collection to a HashSet<string> instead of a List<string> to clarify intent of the collection: order is not a factor, and duplicates are not possible.
- refactored the input validation in the controller using FluentValidation

> - **Benefits:** More readable code, easier to maintain going forward

## ✅ Checklist

### DIFF
--- a/web/src/Web.App/Controllers/LocalAuthorityComparatorsController.cs
+++ b/web/src/Web.App/Controllers/LocalAuthorityComparatorsController.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.AspNetCore.Http.Extensions;
+﻿using FluentValidation;
+using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Mvc;
 using Web.App.Attributes;
 using Web.App.Domain;
@@ -17,7 +18,8 @@ namespace Web.App.Controllers;
 public class LocalAuthorityComparatorsController(
     ILogger<LocalAuthorityComparatorsController> logger,
     IEstablishmentApi establishmentApi,
-    ILocalAuthorityComparatorSetService localAuthorityComparatorSetService)
+    ILocalAuthorityComparatorSetService localAuthorityComparatorSetService,
+    IValidator<LocalAuthorityComparatorSelectionViewModel> validator)
     : Controller
 {
     [HttpGet]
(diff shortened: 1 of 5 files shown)


web/src/Web.App/Controllers/LocalAuthorityComparatorsController.cs
jrabbott, line 1: Nice improvement- I can't help wonder if this could be less painful with a validator..

mattb-hippo: Yeah, that would demonstrate better separation of responsibility. A validator can check for errors leaving the controller to handle actions.  Cleaner.  I'll do that.

mattb-hippo: This is handled in latest commit

web/src/Web.App/Controllers/LocalAuthorityComparatorsController.cs
```diff
@@ -58,41 +60,43 @@ public async Task<IActionResult> Index([FromRoute] string code, [FromQuery] Loca
...
                         break;
                 }
 
-                if (!ModelState.IsValid)
-                {
-                    logger.LogDebug("Posted local authorities comparators failed validation: {ModelState}",
-                        ModelState.Where(m => m.Value != null && m.Value.Errors.Any()).ToJson());
-                }
-                else if (action.Action == FormAction.Continue)
+                if (action.Action == FormAction.Continue)
```
jrabbott: This is really subjective, so keen to get your thoughts. What abut moving the if to above the case? "If then do this, else case and a different return"

```
                if (action.Action == FormAction.Continue)
                {
                    localAuthorityComparatorSetService.SetUserDefinedComparatorSetInSession(code, new UserDefinedLocalAuthorityComparatorSet { Set = comparators.ToArray() });
                    return ContinueActionResult(code, type);    
                }

                switch (action.Action)
                {
                    case FormAction.Add:
                        comparators.Add(viewModel.LaInput!);
                        break;
                    case FormAction.Remove:
                        if (!string.IsNullOrWhiteSpace(action.Identifier))
                        {
                            comparators.Remove(action.Identifier);
                        }
                        break;
                    case FormAction.Reset:
                        comparators = InitialComparatorSetFromNeighbours(localAuthority.StatisticalNeighbours).ToHashSet();
                        break;
                    case FormAction.Clear:
                        comparators = [];
                        break;
                }
                
                return View(nameof(Index), new LocalAuthorityComparatorsViewModel(localAuthority, comparators.ToArray(), type, viewModel.Referrer));
```

jrabbott (dismissed):
Looks so much better, Minor query, but feel free to resolve as it's subjective.

web/src/Web.App/Controllers/LocalAuthorityComparatorsController.cs
mattb-hippo: Yes, I like it.  The final `return View ...` is more connected to the switch, so it makes sense to have the `if action == Continue` bit above that, rather than separating them.  It in the latest commit

jrabbott (approved):

PR #4217, opened 2026-05-21
fix: ensure env vars are loaded into configuration object, and correctly read section and values

jrabbott:
## 🧾 Summary
 This PR standardises how configuration is handled across the Platform APIs. It ensures environment variables are explicitly loaded into the configuration builder and refactors service registration to consistently use the IConfiguration object instead of direct Environment.GetEnvironmentVariable calls.

### 🛠️ Bug Fix Description
- Root Cause: Inconsistencies existed in how environment variables were accessed across different Platform APIs, some were retrieved via IConfiguration while others were accessed directly using Environment.GetEnvironmentVariable. Additionally, some APIs were missing explicit .AddEnvironmentVariables() calls in their configuration builder, potentially leading to missing settings. Configuration was also being accessed using flat keys (e.g., Sql__ConnectionString) rather than the more idiomatic section-based approach.
- Changes:
  - Added .AddEnvironmentVariables() to the HostBuilder configuration in Program.cs across all Platform APIs to ensure environment variables are correctly loaded.
  - Updated Services.Configure methods to pass IConfiguration to sub-registration methods (e.g., AddTelemetry).
  - Refactored extension methods in Platform.Sql, Platform.Cache, and Platform.Search to use the section-based configuration access (e.g., configuration.GetSection("Sql").GetValue<string>("ConnectionString")).
  - Updated packages.lock.json files to ensure consistent dependency states.

## ✅ Checklist
- [x] Code follows project coding standards (e.g.,Trunk-based guidelines, small PR size).
- [x] Tested by running locally (or docs render correctlylocally).
- [x] Relevant documentation updated (if applicable).

## 🔍 Notes
These changes ensure that environment variables using the double-underscore delimiter (e.g., Sql__ConnectionString) are correctly mapped to hierarchical configuration keys (e.g., Sql:ConnectionString). This approach is more robust and provides a consistent interface regardless of whether settings originate from environment variables, appsettings.json, or User Secrets.

### DIFF
--- a/platform/README.md
+++ b/platform/README.md
@@ -78,12 +78,12 @@ The Platform APIs use Azure Functions (Isolated Worker Process), which intention
 The following secrets must be configured for the `platform-local` environment using the `dotnet user-secrets` tool:
 
 ```bash
-dotnet user-secrets set "Sql__ConnectionString" "[value]" --id "platform-local"
-dotnet user-secrets set "Cache__Host" "[value]" --id "platform-local"
-dotnet user-secrets set "Cache__Port" "[value]" --id "platform-local"
-dotnet user-secrets set "Cache__Password" "[value]" --id "platform-local"
-dotnet user-secrets set "Search__Name" "[value]" --id "platform-local"
-dotnet user-secrets set "Search__Key" "[value]" --id "platform-local"
+dotnet user-secrets set "Sql:ConnectionString" "[value]" --id "platform-local"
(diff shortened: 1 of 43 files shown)


J05h-L (dismissed):
The configuration changes make sense and should resolve the build issues we’ve been seeing. After testing locally, I noticed the old flat keys (`Sql__ConnectionString` etc.) no longer resolve. Unless there is something I've overlooked we should update the docs to reflect this (outside of this PR)

J05h-L (approved):

PR #4288, opened 2026-06-10
fix: options form layout

J05h-L:
## 🧾 Summary

QA highlighted layout issues on smaller viewports where elements inside the options form could compress awkwardly. This update improves the responsiveness of the options bar so it wraps cleanly when space becomes constrained.

The options form now supports wrapping, and the checkbox section has been given controlled flex behaviour so labels stay on one line where possible but collapse sensibly when required. This keeps the intended layout without forcing overflow or pushing neighbouring elements out of position.

[AB#298070](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/298070)
[AB#315921](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/315921)

### 🛠️ Bug Fix Description
**Root Cause:**

The previous layout prevented the options from wrapping and restricted checkbox label wrapping, leading to overlap on smaller screens.

**Changes:**

- allow the options form to wrap into multiple rows when needed
- adjust the checkbox container so it can shrink and wrap internally without dominating the row
- maintain sensible spacing and alignment across breakpoints

## ✅ Checklist

## 🔍 Notes

`options-form` is also used on the high needs spending pages. While no issues were previously reported there, the same constraints were present and could cause layout problems at certain resolutions. The updated wrapping behaviour resolves these cases too.

### DIFF
--- a/web/src/Web.App/AssetSrc/scss/main.scss
+++ b/web/src/Web.App/AssetSrc/scss/main.scss
@@ -1241,7 +1241,7 @@ table#current-comparators-la {
 }
 
 .options-form .options-form__checkboxes {
-  flex: 1 1 auto;
+  flex: 0.5 1 auto;
   min-width: 0;
 }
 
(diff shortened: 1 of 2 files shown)


Faizan-ahmad00 (approved):
great. fixes the view as expected.

PR #4298, opened 2026-06-15
chore: update codeowners file to use child team

jrabbott:
## 🧾 Summary

Updates codeowners file to use child team, rather than named people


J05h-L (approved):
Should help route PRs to the right reviewers. I think this will be helpful

PR #4313, opened 2026-06-17
test: add expected fields to API response

murdo-hippo:
## 🧾 Summary
Fix broken platform test from bugfix - the new denominator for percentage income fields now give zeros and are included in API responses instead of being null and being absent from API responses

[AB#316765](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/316765)

### DIFF
--- a/platform/tests/Platform.ApiTests/Data/TrustExpenditureUtilitiesPercentExpenditure.json
+++ b/platform/tests/Platform.ApiTests/Data/TrustExpenditureUtilitiesPercentExpenditure.json
@@ -1,6 +1,10 @@
 {
   "companyNumber": "10192252",
   "trustName": "Test Company/Trust  31",
+  "centralTotalExpenditure": 0.0,
+  "centralTotalUtilitiesCosts": 0.0,
+  "centralEnergyCosts": 0.0,
+  "centralWaterSewerageCosts": 0.0,
   "schoolTotalExpenditure": 100.0,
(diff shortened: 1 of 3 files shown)


Faizan-ahmad00 (dismissed):
This should fix the failing tests. you need to address the automated checks failing on the PR.

J05h-L (commented):
I think adding test as a valid branch name is sensible. Doesn't fit neatly into any of the other categories.

We should update the categories listed in the docs with this change. See `documentation\developers\02_Code-Management.md`

.github/workflows/pr-compliance-checks.yml
```diff
@@ -45,7 +45,7 @@ jobs:
             let currentLabels = pr.labels.map(label => label.name);
 
             // 1. Branch Naming Standards
-            const branchRegex = /^(feature|bugfix|hotfix|exp|tech-debt|docs|prototype|dependabot|chore)(\/[0-9]+)?\/.+$/;
+            const branchRegex = /^(feature|bugfix|hotfix|exp|tech-debt|docs|prototype|dependabot|chore|test)(\/[0-9]+)?\/.+$/;
```
jrabbott: This is a bugfix or hotfix, the regex doesn't need changing.

jrabbott (changes requested):
The regex change needs to be reverted.

murdo-hippo:
see #4314

PR #4314, opened 2026-06-17
fix: Trust percentage income and expenditure test fixes

murdo-hippo:
## 🧾 Summary
Fix broken platform test from bugfix - the new denominator for percentage income fields now give zeros and are included in API responses instead of being null and being absent from API responses

[AB#316765](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/316765)


Faizan-ahmad00 (approved):
Nice. This should fix the failing pipeline.

PR #4315, opened 2026-06-17
fix: Trust percentage income and expenditure test fixes

Faizan-ahmad00:
## 🧾 Summary
Fix broken platform test. 
[AB#316764](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/316764)
[AB#317200](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/317200)

PR #4318, opened 2026-06-18
fix: normalise trust spending only by total trust spending/income

murdo-hippo:
## 🧾 Summary
Another change in [AB#316765](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/316765)
Testing of #4315 revealed a few more places in the database views where expenditure is regularised by eg purely central service income. This corrects them to total service income\expenditure.

There was also a typo in the income regularised field, which was being regularised by expenditure rather than income.


Faizan-ahmad00 (approved):
nice. also fixes the other typo for RR.

PR #4321, opened 2026-06-18
feat: add configuration for secondary region with container app environment

jrabbott:
@mattb-hippo - Adds configuration for secondary region/location with the data-pipeline.

Tested with d11 environment.


mattb-hippo (approved):
Nice

PR #4379, opened 2026-07-23
docs: Update docs to run the pipeline

Faizan-ahmad00:
## 🧾 Summary
the doc mentioned about the queue to trigger the pipeline run which was the old way. updated the doc to align and corrected the queue. 

[AB#313860](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/313860) 
[AB#322819](https://dfe-ssp.visualstudio.com/a14e55df-4fbf-4a2f-a11d-22b187178343/_workitems/edit/322819)

### DIFF
--- a/data-pipeline/README.md
+++ b/data-pipeline/README.md
@@ -86,7 +86,7 @@ These steps will avoid SSL errors due to DfE kit/VPN.
 
 ### Running the pipeline locally
 
-> **Warning:** These instructions run the pipeline **locally only**, against the Azurite emulator. Do not use them to run the pipeline on a deployed environment. To run the pipeline, follow the [Running the pipeline doc](../documentation/data/05_Releases.md), which triggers the run through the `data-pipeline-job-pending` queue and the Orchestrator. Triggering a deployed run by posting directly to `data-pipeline-job-default-start` bypasses the Orchestrator and the run doesn't get tagged with a `JobId` in the database.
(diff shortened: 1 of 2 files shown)


data-pipeline/README.md
```diff
@@ -84,7 +84,9 @@ These steps will avoid SSL errors due to DfE kit/VPN.
 
     * `poetry install`
 
-### Running the pipeline
+### Running the pipeline locally
+
+> **Warning:** These instructions run the pipeline **locally only**, against the Azurite emulator. Do not use them to run the pipeline on a deployed environment. To run the pipeline, follow the [Running the pipeline doc](../documentation/data/05_Releases.md), which triggers the run through the `data-pipeline-job-pending` queue and the Orchestrator. Triggering a deployed run any other way (for example, posting directly to `data-pipeline-job-default-start`) bypasses the Orchestrator and risks the service not operating correctly.
```
murdo-hippo:
```suggestion
> **Warning:** These instructions run the pipeline **locally only**, against the Azurite emulator. Do not use them to run the pipeline on a deployed environment. To run the pipeline, follow the [Running the pipeline doc](../documentation/data/05_Releases.md), which triggers the run through the `data-pipeline-job-pending` queue and the Orchestrator. Triggering a deployed run by posting directly to `data-pipeline-job-default-start` bypasses the Orchestrator and the run doesn't get tagged with a `JobId` in the database.
```
The JobId is used by the frontend to retrieve user-specific custom data runs from the database.

jrabbott: @murdo-hippo the job id isn't used for that purpose. The JobId is used to correlated orchestrations. The front-end does not create the JobId.

The message placed in the pending queue will never have a JobId (default & custom job). It is added by the orchestrator.

jrabbott: @Faizan-ahmad00 you need to revert the change as it's misleading.

Faizan-ahmad00: reverted the changes as suggested and kept it generic rather than specifying the exact issue.

murdo-hippo (approved):

